PR #2, opened 2016-09-17
Fix badges and typo in README.md

stevenmirabito:
Just saw that the badges in the README were a bit broken, so I fixed them, added a missing ")", and cleaned up some of the trailing whitespace. Thanks for making this extension!


rebeckag:
This looks great, thanks for your contribution Steven!

PR #4, opened 2016-09-22
Don't remove values from session while unpacking

stevenmirabito:
After the user logs in, their session is unpacked from the Flask `session` object and copied into the `g` object. However, the values were being popped, leaving an empty session object which Flask helpfully garbage collected. As a result, the user would be redirected to the IdP on every request to the application, which not only added a large number of redirects on every page load but broke any non-GET and XHR requests that the page or application tried to make.

This PR changes the method calls in `_unpack_user_session` to `get()` values from the Flask session object instead of `pop()`ing them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,20 @@ any way with the different client registration modes.
   about the client. The `redirect_uris` registered with the provider MUST include
   `<flask_url>/redirect_uri`, where `<flask_url>` is the URL for the Flask application.
 
+## Configuration
+
 The application using this extension MUST set the following [builtin configuration values of Flask](http://flask.pocoo.org/docs/0.10/config/#builtin-configuration-values):
 
-* `SERVER_NAME` (MUST be the same as `<flask_url>` if using static client registration
-* `SECRET_KEY` (this extension relies on Flask session, which requires `SECRET_KEY`)
+* `SERVER_NAME` (MUST be the same as `<flask_url>` if using static client registration)
+* `SECRET_KEY` (this extension relies on [Flask sessions](http://flask.pocoo.org/docs/0.11/quickstart/#sessions), which requires `SECRET_KEY`)
+
+You may also configure the way Flask sessions handles the user session:
+
+* `PERMANENT_SESSION` (added by this extension; makes the session cookie expire after a configurable length of time instead of being tied to the browser session)
+* `PERMANENT_SESSION_LIFETIME` (the lifetime of a permanent session)
+
+See the [Flask documentation](http://flask.pocoo.org/docs/0.11/config/#builtin-configuration-values) for an exhaustive list of configuration options.
+
+## Example
 
 Have a look at the example Flask app in [app.py](example/app.py) for an idea of how to use it.
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -131,12 +131,11 @@ class OIDCAuthentication(object):
         return wrapper
 
     def _unpack_user_session(self):
-        flask.g.id_token = IdToken().from_dict(flask.session.pop('id_token'))
-        flask.g.access_token = flask.session.pop('access_token', None)
-        userinfo_dict = flask.session.pop('userinfo', None)
+        flask.g.id_token = IdToken().from_dict(flask.session.get('id_token'))
+        flask.g.access_token = flask.session.get('access_token', None)
+        userinfo_dict = flask.session.get('userinfo', None)
         if userinfo_dict:
             flask.g.userinfo = OpenIDSchema().from_dict(userinfo_dict)
-
     def _logout(self):
         id_token_jwt = flask.session['id_token_jwt']
         flask.session.clear()
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -4,8 +4,7 @@ import flask
 from flask.helpers import url_for
 from oic import rndstr
 from oic.oic import Client
-from oic.oic.message import ProviderConfigurationResponse, RegistrationRequest, \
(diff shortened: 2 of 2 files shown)


rebeckag:
Thanks for the contribution!

The `session.pop` is intentional to avoid managing a separate user session in this extension. If we want to do session management separate from whatever session management the IdP does for single-sign on, this needs some more work to do it properly. Some issues to consider:
* How to handle logout? Now we can rely directly on single-sign on from the IdP to handle this for us.
* Is there some mechanism in Flask to automatically expire sessions after a configurable amount of time?

stevenmirabito:
Those are very good points. Flask sessions are [not permanent by default](http://flask.pocoo.org/docs/0.11/api/#flask.session.permanent), so they would expire when the user closes their browser. You could also manage them manually by making them permanent and setting the expire time to some short, fixed (and probably configurable) length of time. Logout could be handled by adding a `/logout` endpoint much like the extension automatically adds the `/redirect_uri` endpoint that would delete the local cookie and optionally redirect to the IdP for single logout.

All that being said, I don't want to reinvent the wheel here if you feel this would be better used in conjunction with something like [Flask-Login](https://github.com/maxcountryman/flask-login) to manage the sessions. Is this extension more of a proof-of-concept or is there a specific way you'd recommend setting this up?

For a bit of context, I'm a member of [Computer Science House](https://github.com/ComputerScienceHouse) and we're moving to OpenID Connect-based SSO with [Keycloak](http://keycloak.org) as our IdP, so I'm currently looking for a solution that I can implement and recommend others migrate their Flask apps to. Thanks for all of your help so far!

rebeckag:
Single logout is a bit complex regarding OpenID Connect. There are a couple of drafts for how to do it, but I currently have no idea what support for example Keycloak (or other providers) has for it. And with that in mind, I don't know how useful local logout is. The user experience will in most cases transparently show the user as logged in the next time they access a protected page since the session with the IdP hasn't been logged out (and SSO will just log them back in).

I've also looked in to Flask-Login, but haven't been able to figure out a reasonable user model since Flask-Login assumes there is some kind of persistent store for user info (which is exactly what we want to avoid with OIDC SSO).

However, there are definitely some improvements we can do in this extension to make the session management a bit more configurable. This PR is already a good start! I have some thoughts on it:
1. We shouldn't unpack the session to `flask.g`, if we still keep the data in the session. So just use for example `flask.session['id_token']`, etc. in the views instead of flask.g.id_token (and the same for the access token and userinfo). Could you update the PR with this change?
2. How do we make the session management as configurable as possible? To replicate the behavior currently implemented it would require the session to not persist between requests, can this be done? Any other behavior should be able to be implemented with Flask default config for sessions (I think..?).

stevenmirabito:
1. I agree that there isn't much point in unpacking the session to `flask.g` if we're not making any changes to the data and accessing, say, the user's name from `session['user_info']['name']` makes more sense to me than having it in a generic global object.
2. If we take advantage of the Flask session module as this extension currently does, this extension would automatically inherit the options that [Flask has built-in](http://flask.pocoo.org/docs/0.11/config/#builtin-configuration-values) for session management. What other options do you think should be configurable? Also, I'm not sure there's a use case for not having the session persist between requests as it causes anything but GET requests to the application to fail, causes AJAX requests to fail, and adds 3 redirects to each request that _does_ go through. In my opinion, that's entirely unacceptable.

As for local logout, I do think it's something that's valuable, and the OIDC spec specifies that the IdP should provide an end session endpoint to the client that can be used to log the user out from the IdP as well. A `/logout` endpoint, in my mind, would delete the local session and redirect to the IdP end session endpoint to end the session at the IdP, then redirect back to the application index. I think whether the IdP supports true "single logout" is irrelevant in the context of this extension.

rebeckag:
1. No, I'd say the three items (ID Token, access token, and Userinfo) currently stored in the session are all logical to expose so we should keep them as is.
2. The use case for not persisting the session between requests comes from the issues with logout I've raised (which were originally pointed out to me by another user) . However, I'm in favor of removing this limitation, and advising setting an extremely short session lifetime instead.

Re logout: The OIDC Core does not publish any endpoints for logout, but I guess you're talking about this: [RP-Initiated Logout](http://openid.net/specs/openid-connect-session-1_0.html#RPLogout)? Something like that shouldn't be too complex, so I'll have a look at implementing support for it.

stevenmirabito:
Yes, that's exactly what I was referring to in terms of logout. I apologize, I wasn't aware that that was part of a seperate draft spec as my IdP supports it. I will take a look at implementing a configurable session lifetime with a fairly short default sometime today and add it to this PR.

rebeckag:
Nice work with the addition of permanent sessions, thanks!
But I can't see that the documented configuration parameter `PERMANENT_SESSION_LIFETIME ` is respected together with `PERMANENT_SESSION`? To get this merged could you add this, as well as remove the unpacking of the session to the global `flask.g` object in  `_unpack_user_session` as we've discussed earlier.

stevenmirabito:
So sorry for the delay in getting this done, and I apologize for just pushing a commit to this PR without commenting. I added the configuration option and removed the unpacking code as requested. As for `PERMANENT_SESSION_LIFETIME`, that's handled by Flask sessions directly and no additional code is required in this extension. See [here](https://github.com/pallets/flask/blob/0664a3f2ddfc227797b5a8f0654e9959b15d9e64/flask/sessions.py#L253) and [here](https://github.com/pallets/flask/blob/0664a3f2ddfc227797b5a8f0654e9959b15d9e64/flask/sessions.py#L358).

rebeckag:
Thank you so much, this looks fantastic!

stevenmirabito:
Thank you for getting this merged so quickly! If everything looks good, would you be able to do a PyPI release with the recent changes?

rebeckag:
Sure thing, v1.0.0 is up on PyPI (and tagged in this repo) now! :)

PR #5, opened 2016-09-22
Pass arguments through the oidc_auth wrapper

stevenmirabito:
The `oidc_auth` decorator did not accept or pass through any arguments to the wrapped route, resulting in an `wrapper() got an unexpected keyword argument` error when the wrapped function expected arguments (e.g. from a variable in the route URL).

This PR changes the `wrapper()` function to both accept any number of arguments and keyword arguments and pass them through to the wrapped function when the user is authenticated, as per the official documentation: http://flask.pocoo.org/docs/0.11/patterns/viewdecorators/


rebeckag:
Thanks, this looks great! :)

Issue #7, opened 2016-10-10
AttributeError: 'IdToken' object has no attribute 'jwt'


stevenmirabito:
It appears that the code for https://github.com/its-dirg/Flask-pyoidc/commit/f86dd443b60a2554602b7f85daa308f267591148 is not working as intended or does not work with some IdPs. On a successful login, I get this stack trace:

```
Traceback (most recent call last):
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 2000, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 1991, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 1567, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/steven/.venv/conditional/lib/python3.5/site-packages/flask_pyoidc/flask_pyoidc.py", line 103, in _handle_authentication_response
    flask.session['id_token_jwt'] = id_token.jwt
AttributeError: 'IdToken' object has no attribute 'jwt'
```

stevenmirabito:
[...]

PR #8, opened 2016-10-11
Convert IdToken to signed JWT for EndSessionRequest

stevenmirabito:
This PR serves as a proposed solution to #7 following the example given in [pyoidc's RP3 sample](https://github.com/rohe/pyoidc/blob/9d186f7a086a76a5a2a09c838dc7be07292eb607/oidc_example/rp3/rp3.py#L431).


stevenmirabito:
Never mind, didn't see that #6 required a newer version of pyoidc. Upgraded the dependency and the problem is resolved.

PR #9, opened 2016-10-12
Add logout view to post_logout_redirect_uris regardless of dynamic registration

stevenmirabito:
This is something I just came across as I don't use dynamic registration. I'm not sure if this is the correct way to do this (you could also say if you're using manual registration you have to register the logout endpoint in `client_registration_info` manually as well), but considering that `logout_view` is registered through the use of the `@oidc_logout` decorator in the application I think it makes sense to register it automatically.


rebeckag:
I actually lean in favour of keeping it as is, due to the first reason you list: if you're using static registration you must explicitly configure what metadata your provider has registered for your client. No exceptions to this (except for `redirect_uris` which is entirely hidden from the user of this extension :)), keeps the behaviour easier to reason about.

PR #22, opened 2017-08-28
Upgrade pyoidc dependency.

zamzterz:


zamzterz:
@andrewkrug Seems pyoidc upgrade was straightforward. 🎉 Thanks for the report!
Have you already tested this with your use case? Please let me know if everything is ok, and I'll release a new version of Flask-pyoidc.

andrewkrug:
@zamzterz to answer your question it would seem that your update here is good.  However, auth0 is not OIDC compliant in their use of the last_updated time stamp and that breaks with the current version of this.  So I can not upgrade until auth0 is OIDC compliant.

PR #25, opened 2018-03-24
handling of nonce in id_token

data-henrik:
The "nonce" attribute is optional and I ran into key errors. The change guards the access and makes it work with my openID Connect provider (IBM Cloud App ID).

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -195,9 +195,11 @@ class OIDCAuthentication(object):
         if 'id_token' in token_resp:
             id_token = token_resp['id_token']
             logger.debug('received id token: %s', id_token.to_json())
-
-            if id_token['nonce'] != flask.session.pop('nonce'):
-                raise ValueError('The \'nonce\' parameter does not match.')
+            
+            # nonce is optional attribute
+            if 'nonce' in id_token:
+               if id_token['nonce'] != flask.session.pop('nonce'):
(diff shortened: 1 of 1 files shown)


zamzterz:
Great catch, thanks for your contribution! 👍 

According to [OIDC spec about nonce in the ID Token](http://openid.net/specs/openid-connect-core-1_0.html#IDToken):
>"If present in the Authentication Request, Authorization Servers MUST include a nonce Claim in the ID Token with the Claim Value being the nonce value sent in the Authentication Request."

So it's not really optional, but since you're working with a provider that doesn't properly support it I think it would make sense to make sending the nonce optional. The current implementation in this PR will silently swallow a missing nonce in the ID Token, which I think is bad.

I think a better option would be the following:
1. Allow setting `extra_request_args['nonce'] = None` in the constructor of `OIDCAuthentication`.
1. In `OIDCAuthentication._authenticate`:  if `extra_request_args['nonce'] is None`, don't generate a nonce and don't set it in `flask.session`.
1. In `OIDCAuthentication._handle_authentication_response`: only verify nonce in the ID Token if `flask.session['nonce']` exists.

Could you try that out and update this PR with some tests for it (otherwise it might take me a while before I get to it)? 🙂

data-henrik:
An update: I contacted the provider who now acknowledged that it needs to be fixed on their end. That's the solution I prefer, so I am going to close thise one.

zamzterz:
Thank you for the update; I agree that's an even better solution! 👍

Issue #44, opened 2019-02-20
problem when behind https terminating proxy


viljoviitanen:
Hello,

I'm running flask-pyoidc behind a https terminating proxy. The app I have works otherwise well, but after login, it issues a redirect to itself as http.

I think the issue is here:

https://github.com/zamzterz/Flask-pyoidc/blob/0dba3ce8931fe5e039c66d7d645331bdbb52960a/src/flask_pyoidc/flask_pyoidc.py#L85

so a request full url is stored (but it's http, as https is terminated elsewhere), and then the redirect back is issued after authentication here:

https://github.com/zamzterz/Flask-pyoidc/blob/0dba3ce8931fe5e039c66d7d645331bdbb52960a/src/flask_pyoidc/flask_pyoidc.py#L148-L149

I have no idea I could convince flask-pyoidc to issue a https redirect here with configuration only, please help?

Otherwise, I think it could be fixed by storing a relative url (i.e. just a path + query parameters). Then whatever configuration present maybe can help flask issue a correct redirect with https. But I have not investigated this option, it's just a thought.

zamzterz:
The HTTP scheme used in the generated redirect_uri is configured via the Flask parameter [`PREFERRED_URL_SCHEME`](http://flask.pocoo.org/docs/1.0/config/#PREFERRED_URL_SCHEME); if you set it to `https` the generated urls should be correct.

viljoviitanen:
Sorry, no. Just setting PREFERRED_URL_SCHEME is not enough here. I have that in my app, and other parts work correctly because of that, but this is not a flask generated url you're redirecting to. Tricks such as setting wsgi.url_scheme do not help here either (they work when flask is generating some urls).

Line 85 stores the full url of the actual request, which really is http://something - because we're running behind https terminating proxy. It does not consider any "standard" such as x-forwarded-proto, and flask cannot do magic when the redirect is issued with a full url. I'm not clever enough with flask to figure out how to force flask.request.url to contain the https url when it really is a http url with some special headers. Perhaps there is a clean way of doing that, I dunno. 

So anyway I ended up doing this... I guess you're not probably gonna accept a PR with this, but I'll share it anyway, maybe someone else finds this useful.
```
diff -ru Flask-pyoidc-2.0.0/src/flask_pyoidc/flask_pyoidc.py Flask-pyoidc-2.0.0-new/src/flask_pyoidc/flask_pyoidc.py
[...]

PR #45, opened 2019-05-07
Djbeadle https redirect uri

djbeadle:
I've been working with Azure App service which does not allow **_any_** http callbacks except for localhost, so the ProxyFix solution mentioned in #44 doesn't apply.

This change sets the parameter "_scheme" on url_for() in OIDCAuthentication.init_app() based on the config value "REDIRECT_URI_FORCE_HTTPS". If the config value is set to True it the redirect uri will be configured to use https. 

If no value is set or REDIRECT_URI_FORCE_HTTPS equals false the redirect uri equals http and no behavior is changed.

I'm not sure where to document this setting, but if accepted please let me know and I will write it up!

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -62,7 +62,7 @@ class OIDCAuthentication(object):
         with app.app_context():
             url_scheme = 'https' if current_app.config.get('REDIRECT_URI_FORCE_HTTPS', False) else 'http'
             self.clients = {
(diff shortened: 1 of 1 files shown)


zamzterz:
Thanks for the contribution! 🎉 
What is the use-case for which the builtin Flask configuration value [`PREFERRED_URL_SCHEME`](http://flask.pocoo.org/docs/1.0/config/#PREFERRED_URL_SCHEME) doesn't work?

djbeadle:
Whoops, I somehow missed that. It's a much better solution!

PR #57, opened 2019-07-27
Add ability to configure redirect uri (fixes #51)

### LINKED ISSUE #51: Do not hardcode routes like `/redirect_uri`
Better provide a default which people can then override.

rgmz:
Presently this is missing support for the implicit flow:

https://github.com/zamzterz/Flask-pyoidc/blob/c1494d755af1d427bc214446699eb97f951eb09d/src/flask_pyoidc/files/parse_fragment.html#L5

https://github.com/zamzterz/Flask-pyoidc/blob/c1494d755af1d427bc214446699eb97f951eb09d/src/flask_pyoidc/flask_pyoidc.py#L164

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -54,17 +54,13 @@ class OIDCAuthentication(object):
         self.clients = None
         self._logout_view = None
         self._error_view = None
-        self._redirect_uri_endpoint = 'redirect_uri'
+        self._redirect_uri_endpoint = None
 
         if app:
             self.init_app(app)
 
     def init_app(self, app):
-        if 'OIDC_REDIRECT_ENDPOINT' in app.config:
-            redirect_endpoint = app.config.get('OIDC_REDIRECT_ENDPOINT')
-            if redirect_endpoint[0] == '/':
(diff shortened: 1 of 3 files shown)


src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -55,21 +54,28 @@ def __init__(self, provider_configurations, app=None):
...
         self._error_view = None
+        self._redirect_uri_endpoint = 'redirect_uri'
 
         if app:
             self.init_app(app)
 
     def init_app(self, app):
+        if 'OIDC_REDIRECT_ENDPOINT' in app.config:
+            redirect_endpoint = app.config.get('OIDC_REDIRECT_ENDPOINT')
+            if redirect_endpoint[0] == '/':
```
zamzterz: I think this can be more permissive and allow the leading `/`, simply removing it if it's there. Then reading the value from config can also be simplified as:
```python
self._redirect_uri_endpoint = app.config.get('OIDC_REDIRECT_ENDPOINT', 'redirect_uri').lstrip('/')
```

With this the initialisation in the constructor should be changed to `self._redirect_uri_endpoint = None` to avoid duplicating the default value.

Issue #46, opened 2019-05-16
Uninitialised session error


rhenriquez28:
Hi!

I followed the example app and I've been able to connect to my auth endpoint. However, after I log in with the user and password I get a 500 error and the terminal says:

```
File "/usr/local/lib/python3.6/site-packages/flask_pyoidc/flask_pyoidc.py", line 101, in _handle_authentication_response
    client = self.clients[UserSession(flask.session).current_provider]
  File "/usr/local/lib/python3.6/site-packages/flask_pyoidc/user_session.py", line 19, in __init__
    raise UninitialisedSession("Trying to pick-up uninitialised session without specifying 'provider_name'")
```

I notice that the reason for this error is that since the library is calling UserSession without the provider_name parameter the first condition fails and it raises the error.

How can I fix this?

zamzterz:
The `provider_name` should already be set internally when using the `oidc_auth` decorator function.
The error you get might indicate the Flask session isn't correctly included in the request back to the redirect URI. Could you verify whether that's the case or provide some more detail about how your implementation looks (i.e. a minimal working example)?

rhenriquez28:
Sorry, I decided to use another way to connect to the auth endpoint since I was in a hurry and it worked. Thanks for replying so quickly

enkelli:
Hi, I get the same error when I try to logout twice subsequently. First logout passes without errors, second call fails. (I mean login -> logout -> logout, not login -> logout -> login -> logout)
I know it is not common to hit "logout" endpoint multiple times but I would not expect server error when trying to logout when no user is logged-in.

zamzterz:
@enkelli Nice catch, this has been fixed in #69. 🙂

rokpu:
Hey, so I'm running into the same problem. At first, it redirects correctly, but when I log in, I get the 500 error. 

Here's some code
```
app.config.update(OIDC_REDIRECT_URI='xxxxxxx', SECRET_KEY='xxxxxx')

client_id = "test-client-id"
client_secret = "test-client-secret"
client_metadata = ClientMetadata(client_id=client_id, client_secret=client_secret)
config = ProviderConfiguration(
  issuer='id-provider',
  client_metadata=client_metadata,
  auth_request_params={'scope': ['openid', 'profile', 'email']}
)
auth = OIDCAuthentication({'provider_name': config}, app)

@app.route('/login')
@auth.oidc_auth('provider_name')
def login():
  return redirect('/test')

@app.route('/test')
[...]

PR #69, opened 2019-12-10
Handle logout gracefully when no active user session.

zamzterz:
As reported [here](https://github.com/zamzterz/Flask-pyoidc/issues/46#issuecomment-563238184).


enkelli:
Great, thank you! :+1:

PR #73, opened 2020-02-13
Forward ssl verify to oic

MrJacek:
Hello,
I not found how to forward ssl_verify to oic objects. I thing it was missing so i add this in PR. 
Please review this small change :)

All the best.
MrJacek

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+venv/*
 *.pyc
 *.egg-info
 build/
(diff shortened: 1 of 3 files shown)


zamzterz:
Thanks for your contribution! 👏 

I'd rather not expose this option directly in Flask-pyoidc to encourage proper SSL/TLS usage.

The underlying library (pyoidc) is using `requests` under the hood, so if you're using a self-signed certificate you can configure (`REQUESTS_CA_BUNDLE`)[https://2.python-requests.org/en/master/user/advanced/#ssl-cert-verification] to make it trusted instead.
Let me know if that solves your use case!

PR #104, opened 2021-06-11
fix: session expire date parsing

Titotix:


zamzterz:
Thanks for your contribution - great start! 👍 

As mentioned in #106, it would be fine to drop testing against Python 3.5 in this PR. And it would be great to also combine it with #105 to get the tests passing again with a single PR.

Titotix:
I had a commit to drop Python3.5 support.
But this #105 doesn't fix anything actually, it was to test if the upgrade of pip was enough to fix the missing Rust compiler for cryptography lib but it is not and I did not dig more in it. According to the cryptography doc :

> If you are on RHEL/CentOS/Fedora/Debian/Ubuntu or another distribution derived from the preceding list, then you should upgrade pip (in a virtual environment!) and attempt to install cryptography again before trying to install the Rust toolchain. These platforms will receive a binary wheel and require no compiler if you have an updated pip!

But I am not used to travis and I didn't look for which platform it uses to run the test.

zamzterz:
After a closer look, and combining your 2 PRs with a [small addition](https://github.com/zamzterz/Flask-pyoidc/pull/107/commits/a90573803d350672e984ef3779e69eedcfec9558) the tests are now working again.
I took the liberty of creating a PR out of that which I'd be happy if you took a quick look at: #107

PR #107, opened 2021-06-16
Test fixes

zamzterz:
Tests were failing to install 'cryptography' dependency on Travis and a small change in date format was introduced in Flask.

This also enables running the tests for Python 3.9 to ensure compatibility, while dropping testing agains Python 3.5 (which has reached EOL).

_Closes #106_


Titotix:
Great, sounds good :)

PR #111, opened 2021-10-16
Fix deprecated collections import for python 3.10

ThoreKr:
When trying to include this in a python 3.10 environment I got the following traceback:

```
   File "/usr/local/lib/python3.10/site-packages/flask_pyoidc/provider_configuration.py", line 9, in <module>
     class OIDCData(collections.MutableMapping):
 AttributeError: module 'collections' has no attribute 'MutableMapping'
 ```
 
 Apparently `collections.MutableMapping` has been deprecated in 3.9 and replaced in 3.10 with `collections.abc.MutableMapping` (at least that is what stackoverflow claimed).

This patch swaps them out and extends tox to test python 3.10 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 
 matrix:
   include:
-    - python: 3.10
+    - python: 3.10-dev
       env: TOXENV=py310
     - python: 3.9
       env: TOXENV=py39


ThoreKr:
Thanks for fixing the remaining bit.
For unknown reasons github didn't send out any emails

Issue #116, opened 2021-12-11
Cannot build url for 'logout' when using dynamic client registration


infohash:
> If the logout view is mounted under a custom endpoint (other than the default, which is the name of the view function), or if using Blueprints, you must specify the full URL in the Flask-pyoidc configuration using `post_logout_redirect_uris`.

This is working as intended but when using dynamic client registration these methods are called:

https://github.com/zamzterz/Flask-pyoidc/blob/b0058c0724f35d6999322d0a584003185c3189d9/src/flask_pyoidc/flask_pyoidc.py#L87

https://github.com/zamzterz/Flask-pyoidc/blob/b0058c0724f35d6999322d0a584003185c3189d9/src/flask_pyoidc/flask_pyoidc.py#L95-L97

https://github.com/zamzterz/Flask-pyoidc/blob/b0058c0724f35d6999322d0a584003185c3189d9/src/flask_pyoidc/flask_pyoidc.py#L89

https://github.com/zamzterz/Flask-pyoidc/blob/b0058c0724f35d6999322d0a584003185c3189d9/src/flask_pyoidc/flask_pyoidc.py#L84-L85

`self._logout_view.__name__` returns the name of the view function. `url_for` uses the route registration to find the route registered against that name. If the view function is mounted at blueprint, its name is prefixed by the name of the blueprint so providing only `__name__` to `url_for` is not enough.

    blueprint = Blueprint('api', __name__, url_prefix='/api/v1')
    @blueprint.route('/logout')
    def logout():

        return 'You have been logged out'

    print(url_for('api.logout'))
    # http://localhost:5000/api/v1/logout
    print(url_for('logout'))
    # Cannot build url for 'logout'. Did you mean 'api.logout'?

`.get()` will always call that method to set default value even if `post_logout_redirect_uris` is provided as an argument.

PR #117, opened 2021-12-11
fixed url_for build for post_logout_redirect_uris

infohash:
Addresses #116 

Fixed url_for build error of post_logout_redirect_urls when using dynamic client registration

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -88,6 +88,7 @@ class OIDCAuthentication:
                 return url_for(self._logout_view.__name__, _external=True)
             except BuildError:
                 logger.info('logout view is mounted under a custom endpoint')
+                raise
 
     def _register_client(self, client):
         def default_post_logout_redirect_uris():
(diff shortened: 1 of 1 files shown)


zamzterz:
Thanks for the great contribution! 👏 

As you point out in #116 the default value is incorrectly calculated when `post_logout_redirect_uris` is already set here:
https://github.com/zamzterz/Flask-pyoidc/blob/b0058c0724f35d6999322d0a584003185c3189d9/src/flask_pyoidc/flask_pyoidc.py#L95-L97

I think it would be more robust to fix that instead, i.e. `default_post_logout_redirect_uris()` should only be called when the `get()` returns `None`.
That way, the `BuildError` is not "silently" swallowed, hiding possible misconfiguration. WDYT?

infohash:
Agreed. If get returns None then only `default_post_logout_redirect_uris()` will be raised. It will raise an exception if the logout endpoint is mounted at custom endpoint and user doesn't provide it as an argument.

zamzterz:
Sounds good then! Would you be able to update this PR to reflect that change instead?

zamzterz:
@infohash Thanks for the swift changes. 👍 

I just added some minor formatting changes, and will release a new version with this fix in just a bit.

Issue #75, opened 2020-04-13
Decorator for validating token


npcode15:
Hello,

I would like to know if there is a separate decorator for validating the token?
or is the use of @auth.oidc_auth dual purpose?

Login and logout with browser works perfectly however, I am not able to test the API using Postman.
Steps

Logged in (using browser), received the token
Copied the token to Postman 'Auth' to be used as a 'Bearer'
Called a resource (API) with '@auth.oidc_auth(default)' decorator - Postman returned the HTML page as it requires the login process again (probably discarded the access token)
Please let me know.

Thank you.

npcode15:
@zamzterz Thanks for the response. Can I take this task?

zamzterz:
Please do! And let me know if you need any help.

npcode15:
@zamzterz will do. Thanks!

zamzterz:
Closing this due to inactivity, please re-open if the use case applies.

infohash:
I'm in the same situation right now. I've REST API endpoints that I've decorated with @auth.oidc_auth('default') using Authorization Code Flow. But another microservice cannot access the endpoints because it redirects them with 302 to Identity Provider Authorization Endpoint.

Now this all looks good when you are accessing the endpoints using browser in which the user can simply sign in to Identity Provider but microservices won't understand that.

Is there a way to streamline this process without browser. Identity Providers like Keycloak allows sign in using API also. So instead of visiting URL in browser, it takes username, password, scope, redirect_uri, state and optionally nonce in the API call and returns the authorization code.

zamzterz:
@infohash That sounds like the resource owner password flow which is not recommended in general deployments (see for example [here](https://www.scottbrady91.com/oauth/why-the-resource-owner-password-credentials-grant-type-is-not-authentication-nor-suitable-for-modern-applications)), hence it's not supported in this extension.

If you really need to use it I recommend you implement it outside this extension.
However, you could potentially use some of the non-public methods from this extension to help you:
```python
auth.init_app(app) # OIDCAuthentication must be initialised before

# client_id and client_secret is automatically injected
token_req = {
    'grant_type': 'password',
    'username': username,
    'password': password,
    'scope': scope
}
token_response = auth.clients['default']._token_request(token_request)
```
or use the undelrying library [`pyoidc`](https://github.com/OpenIDC/pyoidc) directly.

infohash:
[...]

PR #120, opened 2021-12-19
added token introspection and client credentials flow

infohash:
## Changelogs:

Addresses #75

## Token Introspection

Now the user can send API request directly from curl or from REST API clients. Token introspection method verifies the token before allowing access.

## Token Based Authorization

@auth.token_auth decorator is added to support token based authorization. Token based authorization is backed by token introspection.

## Dual Purpose

@auth.access_control decorator is added that serves dual purpose, that is it can authenticate the user as well as validate the token.

How to use these features is added in its docstring.

## Client Credentials Flow

The service can use Client Credentials Flow to request access token by calling a new public method. With machine-to-machine (M2M) applications, such as CLIs, daemons, or services running on your back-end, the system authenticates and authorizes the app rather than a user. M2M apps use the Client Credentials Flow (defined in OAuth 2.0 RFC 6749, section 4.4), in which they pass along their Client ID and Client Secret to authenticate themselves and get a token. This also eliminates the necessity of having a user-agent in order to authenticate with the Identity Provider. Other services can also use this method to request access token in order to access your APIs.

Client Credentials is backed by token introspection so any usage of access token obtained by client credential flow is verified by token introspection.

How to use this method is added in its docstring.

---

This update doesn't introduce any breaking changes to the codebase of the user. It only adds optional arguments which are default to false should the user choose to update the extension.

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -17,18 +17,19 @@ import functools
 import json
 import logging
 import time
+from typing import Union
 from urllib.parse import parse_qsl
 
 import flask
 import importlib_resources
-from flask import current_app, g
+from flask import _app_ctx_stack, current_app
 from flask.helpers import url_for
 from oic import rndstr
-from oic.exception import AccessDenied, NotForMe
 from oic.extension.message import TokenIntrospectionResponse
 from oic.oic import AuthorizationRequest
 from oic.oic.message import EndSessionRequest
 from werkzeug.exceptions import HTTPException
+from werkzeug.local import LocalProxy
 from werkzeug.routing import BuildError
 from werkzeug.utils import redirect
 
@@ -46,8 +47,7 @@ class OIDCAuthentication:
     """
 
     def __init__(self, provider_configurations, app=None,
-                 redirect_uri_config=None,
-                 disable_authorization=False):
+                 redirect_uri_config=None):
         """
         Args:
             provider_configurations (Mapping[str, ProviderConfiguration]):
@@ -61,14 +61,14 @@ class OIDCAuthentication:
         self.clients = None
         self._logout_view = None
         self._error_view = None
+        # current_identity proxy to obtain user info whose token was passed in
+        # the request. It is available until current request only and is
+        # destroyed between the request. The value is set by token_auth
+        # decorator.
+        self.current_identity = LocalProxy(lambda: getattr(_app_ctx_stack.top,
+                                                           'current_identity',
+                                                           None))
         self._redirect_uri_config = redirect_uri_config
-        # This is a lazy way to disable authorization globally. If
-        # self.disable_authorization is set to True, it disables authorization
-        # for all API endpoints that are decorated with @oidc_and_token_auth
-        # and is False by default. This is helpful when you want to deprecate
-        # developer access to your APIs that work with both oidc and token
-        # authorization.
-        self.disable_authorization = disable_authorization
 
         if app:
             self.init_app(app)
@@ -350,7 +350,6 @@ class OIDCAuthentication:
         '''
         if 'Authorization' in request.headers and request.headers['Authorization'].startswith('Bearer '):
             return True
-        logger.info('missing authorization field')
         return False
 
     def _parse_access_token(self, request) -> str:
@@ -369,7 +368,8 @@ class OIDCAuthentication:
         _, access_token = request.headers['Authorization'].split(maxsplit=1)
         return access_token
 
-    def introspect_token(self, request, client, scopes: list = None) -> TokenIntrospectionResponse:
+    def introspect_token(self, request, client, scopes: list = None) -> Union[
+            TokenIntrospectionResponse, False]:
         '''RFC 7662: Token Introspection
         The Token Introspection extension defines a mechanism for resource
         servers to obtain information about access tokens. With this spec,
@@ -407,15 +407,16 @@ class OIDCAuthentication:
             return False
         # Check if client_id is in audience claim
         if not client._client.client_id in result['aud']:
+            # log the exception if client_id is not in audience and returns
+            # False, you can configure audience with Identity Provider
             logger.info('Token is valid but required audience is missing')
-            # raises exception if client_id is not in audience, you can
-            # configure audience from Identity Provider
-            raise NotForMe('required audience is missing')
-        # Check if scopes associated with the access_token are the ones
-        # given by the user and not something else which is not permitted.
+            return False
+        # Check if the scopes associated with the access_token are the ones
+        # required by the endpoint and not something else which is not
+        # permitted.
         if scopes and not set(scopes).issubset(set(result['scope'])):
             logger.info('Token is valid but does not have required scopes')
-            raise NotForMe('Token does not have required scopes')
+            return False
         return result
 
     def token_auth(self, provider_name, scopes_required: list = None):
@@ -430,15 +431,14 @@ class OIDCAuthentication:
 
         Raises
         ------
-        AccessDenied
+        flask.abort(401)
             If authorization field is missing.
-        HTTPException
+        flask.abort(403)
             If the access token is invalid.
 
         How To Use
         ----------
-        >>> auth = OIDCAuthentication({'default': provider_config},
-                                      disable_authorization=False)
+        >>> auth = OIDCAuthentication({'default': provider_config})
         >>> @app.route('/')
             @auth.token_auth(provider_name='default')
             def index():
@@ -455,7 +455,9 @@ class OIDCAuthentication:
                 client = self.clients[provider_name]
                 # Check for authorization field in the request header.
                 if not self._check_authorization_header(flask.request):
-                    raise AccessDenied('missing authorization field')
+                    logger.info('Request header has no authorization field')
+                    # Abort the request if authorization field is missing.
+                    flask.abort(401)
                 token_introspection_result = self.introspect_token(
                     request=flask.request, client=client,
                     scopes=scopes_required)
@@ -463,10 +465,10 @@ class OIDCAuthentication:
                     logger.info('user has valid access token')
                     # Store token introspection info within the application
                     # context.
-                    g._token_introspection_info = token_introspection_result.to_dict()
+                    _app_ctx_stack.top.current_identity = token_introspection_result.to_dict()
                     return view_func(*args, **kwargs)
                 # Forbid access if the access token is invalid.
-                flask.abort(401)
+                flask.abort(403)
 
             return wrapper
 
@@ -493,8 +495,7 @@ class OIDCAuthentication:
 
         How To Use
         ----------
-        >>> auth = OIDCAuthentication({'default': provider_config},
-                                      disable_authorization=False)
+        >>> auth = OIDCAuthentication({'default': provider_config})
         >>> @app.route('/')
             @auth.access_control(provider_name='default')
             def index():
@@ -502,13 +503,8 @@ class OIDCAuthentication:
         >>> # You can also specify scopes required by your endpoint.
         >>> @auth.access_control(provider_name='default',
                                  scopes_required=['read', 'write'])
-        >>> # Additionally, you can disable token authorization entirely by
-            # setting disable_authorization to True. It is helpful when you
-            # want to deprecate developer access to your APIs.
-        >>> auth = OIDCAuthentication({'default': provider_config},
-                                      disable_authorization=True)
         '''
-        def combined_decorator(view_func):
+        def hybrid_decorator(view_func):
 
             fallback_to_oidc = self.oidc_auth(provider_name)(view_func)
 
@@ -516,23 +512,31 @@ class OIDCAuthentication:
             def wrapper(*args, **kwargs):
 
                 try:
-                    if self.disable_authorization:
-                        logger.info('Authorization is disabled, this endpoint can use OIDC only')
-                        flask.abort(403)
-                    return self.token_auth(provider_name,
-                                           scopes_required)(
-                                               view_func)(*args, **kwargs)
+                    # Check if the request contains cookies. It's possible that
+                    # a request contains both token and cookies. If it does
+                    # then instead of checking for token, let oidc_auth handles
+                    # the request because cookies are not usually sent by the
+                    # rest api client. If cookies are present that means
+                    # there's some user-agent (browser) involved which should
+                    # be handled by oidc_auth.
+                    if not flask.request.cookies:
+                        return self.token_auth(provider_name,
+                                               scopes_required)(view_func)(
+                                                   *args, **kwargs)
+                    return fallback_to_oidc(*args, **kwargs)
                 except HTTPException as ex:
+                    # token_auth will raise the HTTPException if either
+                    # authorization field is missing from the request header or
+                    # token is invalid. If the authorization field is missing,
+                    # fallback to oidc.
                     logger.debug(ex.code)
-                    if ex.code == 401:
-                        flask.abort(401)
+                    if ex.code == 403:
+                        # If token is present but it's invalid, do not fallback
+                        # to oidc_auth. Instead, abort the request.
+                        flask.abort(403)
                     return fallback_to_oidc(*args, **kwargs)
-                except AccessDenied:
-                    return fallback_to_oidc(*args, **kwargs)
-                except NotForMe as ex:
-                    flask.abort(403, ex.args)
 
             return wrapper
 
-        return combined_decorator
+        return hybrid_decorator
 
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -334,8 +334,9 @@ class OIDCAuthentication:
                        refresh_token=response.get('refresh_token'))
         return access_token
 
-    def _check_authorization_header(self, request) -> bool:
-        '''Look for authorization in request header.
+    @staticmethod
+    def _check_authorization_header(request) -> bool:
+        """Look for authorization in request header.
 
         Parameters
         ----------
@@ -346,13 +347,14 @@ class OIDCAuthentication:
         -------
         bool
             True if the request header contains authorization else False.
-        '''
+        """
         if 'Authorization' in request.headers and request.headers['Authorization'].startswith('Bearer '):
             return True
         return False
 
-    def _parse_access_token(self, request) -> str:
-        '''Parse access token from the authorization request header.
+    @staticmethod
+    def _parse_access_token(request) -> str:
+        """Parse access token from the authorization request header.
 
         Parameters
         ----------
@@ -363,13 +365,13 @@ class OIDCAuthentication:
         -------
         accept_token : str
             access token from the request header.
-        '''
+        """
         _, access_token = request.headers['Authorization'].split(maxsplit=1)
         return access_token
 
     def introspect_token(self, request, client, scopes: list = None) -> Union[
             TokenIntrospectionResponse, bool]:
-        '''RFC 7662: Token Introspection
+        """RFC 7662: Token Introspection
         The Token Introspection extension defines a mechanism for resource
         servers to obtain information about access tokens. With this spec,
         resource servers can check the validity of access tokens, and find out
@@ -381,7 +383,7 @@ class OIDCAuthentication:
         request : werkzeug.local.LocalProxy
             flask request object.
         client : flask_pyoidc.pyoidc_facade.PyoidcFacade
-            PyoidcFacdae object contains metadata of the provider and client.
+            PyoidcFacade object contains metadata of the provider and client.
         scopes : list
             Specify scopes that are bound for the end endpoint.
 
@@ -395,7 +397,7 @@ class OIDCAuthentication:
         NotForMe
             If access_token is invalid.
 
-        '''
+        """
         received_access_token = self._parse_access_token(request)
         # send token introspection request
         result = client._token_introspection_request(
@@ -405,7 +407,7 @@ class OIDCAuthentication:
         if not result.get('active'):
             return False
         # Check if client_id is in audience claim
-        if not client._client.client_id in result['aud']:
+        if client._client.client_id not in result['aud']:
             # log the exception if client_id is not in audience and returns
             # False, you can configure audience with Identity Provider
             logger.info('Token is valid but required audience is missing')
@@ -419,7 +421,7 @@ class OIDCAuthentication:
         return result
 
     def token_auth(self, provider_name, scopes_required: list = None):
-        '''Token based authorization.
+        """Token based authorization.
 
         Parameters
         ----------
@@ -445,7 +447,7 @@ class OIDCAuthentication:
         >>> # You can also specify scopes required by your endpoint.
         >>> @auth.token_auth(provider_name='default',
                              scopes_required=['read', 'write'])
-        '''
+        """
         def token_decorator(view_func):
 
             @functools.wraps(view_func)
@@ -475,7 +477,7 @@ class OIDCAuthentication:
 
     def access_control(self, provider_name: str,
                        scopes_required: list = None):
-        '''This decorator serves dual purpose that is it can do both token
+        """This decorator serves dual purpose that is it can do both token
         based authorization and oidc based authentication. If your API needs
         to be accessible by either modes, use this decorator otherwise use
         either oidc_auth or token_auth.
@@ -502,7 +504,7 @@ class OIDCAuthentication:
         >>> # You can also specify scopes required by your endpoint.
         >>> @auth.access_control(provider_name='default',
                                  scopes_required=['read', 'write'])
-        '''
+        """
         def hybrid_decorator(view_func):
 
             fallback_to_oidc = self.oidc_auth(provider_name)(view_func)
@@ -511,18 +513,11 @@ class OIDCAuthentication:
             def wrapper(*args, **kwargs):
 
                 try:
-                    # Check if the request contains cookies. It's possible that
-                    # a request contains both token and cookies. If it does
-                    # then instead of checking for token, let oidc_auth handles
-                    # the request because cookies are not usually sent by the
-                    # rest api client. If cookies are present that means
-                    # there's some user-agent (browser) involved which should
-                    # be handled by oidc_auth.
-                    if not flask.request.cookies:
-                        return self.token_auth(provider_name,
-                                               scopes_required)(view_func)(
-                                                   *args, **kwargs)
-                    return fallback_to_oidc(*args, **kwargs)
+                    # If the request header contains auhorization, token_auth
+                    # verifies the access_token otherwise an exception occurs
+                    # and the request falls back to oidc_auth.
+                    return self.token_auth(provider_name, scopes_required)(
+                        view_func)(*args, **kwargs)
                 except HTTPException as ex:
                     # token_auth will raise the HTTPException if either
                     # authorization field is missing from the request header or
@@ -530,12 +525,11 @@ class OIDCAuthentication:
                     # fallback to oidc.
                     logger.debug(ex.code)
                     if ex.code == 403:
-                        # If token is present but it's invalid, do not fallback
-                        # to oidc_auth. Instead, abort the request.
+                        # If token is present, but it's invalid, do not fall
+                        # back to oidc_auth. Instead, abort the request.
                         flask.abort(403)
                     return fallback_to_oidc(*args, **kwargs)
 
             return wrapper
 
         return hybrid_decorator
-
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -21,10 +21,11 @@ from urllib.parse import parse_qsl
 
 import flask
 import importlib_resources
-from flask import current_app
+from flask import current_app, g
 from flask.helpers import url_for
 from oic import rndstr
-from oic.exception import NotForMe
+from oic.exception import AccessDenied, NotForMe
+from oic.extension.message import TokenIntrospectionResponse
 from oic.oic import AuthorizationRequest
 from oic.oic.message import EndSessionRequest
 from werkzeug.routing import BuildError
@@ -207,7 +208,7 @@ class OIDCAuthentication:
         return 'Something went wrong with the authentication, please try to login again.'
 
     def oidc_auth(self, provider_name, accept_token: bool = None,
-                  scopes_required : list = None):
+                  scopes_required: list = None):
         if provider_name not in self._provider_configurations:
             raise ValueError(
                 "Provider name '{}' not in configured providers: {}.".format(provider_name,
@@ -237,10 +238,16 @@ class OIDCAuthentication:
                 # passed access token in the request header and make
                 # introspection request to the identity provider to verify the
                 # token.
-                elif accept_token and self._check_authorization_header(flask.request):
-                    if self.introspect_token(request=flask.request,
-                                             client=client, scopes=scopes_required):
+                elif accept_token and self._check_authorization_header(
+                     flask.request):
+                    token_introspection_result = self.introspect_token(
+                        request=flask.request, client=client,
+                        scopes=scopes_required)
+                    if token_introspection_result:
                         logger.info('user has valid access token')
+                        # Store token introspection info within the application
+                        # context.
+                        g.token_introspection_info = token_introspection_result.to_dict()
                         return view_func(*args, **kwargs)
                     # Forbid access if the access token is invalid.
                     flask.abort(403)
@@ -358,6 +365,7 @@ class OIDCAuthentication:
         '''
         if 'Authorization' in request.headers and request.headers['Authorization'].startswith('Bearer '):
             return True
+        logger.info('missing authorization field')
         return False
 
     def _parse_access_token(self, request) -> str:
@@ -376,7 +384,7 @@ class OIDCAuthentication:
         _, access_token = request.headers['Authorization'].split(maxsplit=1)
         return access_token
 
-    def introspect_token(self, request, client, scopes: list = None) -> bool:
+    def introspect_token(self, request, client, scopes: list = None) -> TokenIntrospectionResponse:
         '''RFC 7662: Token Introspection
         The Token Introspection extension defines a mechanism for resource
         servers to obtain information about access tokens. With this spec,
@@ -411,8 +419,8 @@ class OIDCAuthentication:
                                       access_token_required=True)
 
         >>> # You can also enable or disable it per API endpoint from the
-        # decorator by providing accept_token to either True or False as an
-        # argument. It has higher precedence than global argument.
+            # decorator by providing accept_token to either True or False as an
+            # argument. It has higher precedence than global argument.
 
         >>> @auth.oidc_auth('default', accept_token=True)
         >>> # Additionally, you can provide required scopes for your endpoint.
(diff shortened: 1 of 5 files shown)


src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -203,12 +206,20 @@ def _show_error_response(self, error_response):
 
         return 'Something went wrong with the authentication, please try to login again.'
 
-    def oidc_auth(self, provider_name):
+    def oidc_auth(self, provider_name, accept_token: bool = None,
```
zamzterz, line 209: I think this should be a separate decorator function to avoid mixing authorization methods.
Either an endpoint should rely on the user session stored in cookies, or it should accept access tokens.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -222,6 +233,18 @@ def wrapper(*args, **kwargs):
...
                     return view_func(*args, **kwargs)
+                # If the user is not authenticated, check if the user has
+                # passed access token in the request header and make
+                # introspection request to the identity provider to verify the
+                # token.
+                elif accept_token and self._check_authorization_header(flask.request):
+                    if self.introspect_token(request=flask.request,
+                                             client=client, scopes=scopes_required):
+                        logger.info('user has valid access token')
+                        return view_func(*args, **kwargs)
```
zamzterz, line 244: This should pass some of the introspected token data to the `view_func`, like the `sub` claim if it's in the token and any client_id, to allow the `view_func` to determine which client/user is calling the endpoint.

This would aslo be yet another reason to have it as separate decorator, since it would pass som arguments to the decorated view function.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -319,3 +342,100 @@ def valid_access_token(self, force_refresh=False):
...
+         request : werkzeug.local.LocalProxy
+             flask request object
+             
+         Returns
+         -------
+         bool
+             True if the request header contains authorization.
+         '''
+         if 'Authorization' in request.headers and request.headers['Authorization'].startswith('Bearer '):
+             return True
```
zamzterz, line 360: Missing `return` for the negative case.
(It will return `None` default instead of `false`.)

This could potentially be combined with `_parse_access_token`:
if a token exists in the `Authorization` header, parse it from there, otherwise return `None`.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -319,3 +342,100 @@ def valid_access_token(self, force_refresh=False):
...
+                # raises exception if client_id is not in audience, you can
+                # configure audience from Identity Provider
+                raise NotForMe('required audience is missing')
+            # Check if scopes associated with the access_token are the ones
+            # given by the user and not something else which is not permitted.
+            if scopes is not None:
+                if not set(scopes).issubset(set(result['scope'].split())):
+                    logger.info('Token is valid but does not have required scopes')
+                    raise NotForMe('Token does not have required scopes')
+            return True
```
zamzterz, line 441: * Missing return for the negative case (no `return False`).
  The deep nesting could also be avoided by reversing the statements, i.e.:

   ```python
   if not result.get('active'):
     return False

   if not client._client.client_id in result['aud']:
     logger.info('Token is valid but required audience is missing')
     raise NotForMe('required audience is missing')

   etc...
   ```
* As mentioned in a comment above, some of the data in the Introspection response should be passed to the view function. Hence this function needs to return either the IntrospectionResponse or a subset of its data instead of a boolean.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -50,6 +51,9 @@ def __init__(self, provider_configuration, redirect_uri):
         """
         self._provider_configuration = provider_configuration
         self._client = Client(client_authn_method=CLIENT_AUTHN_METHOD)
+        # Token Introspection is implmented in extension sub-package of the
+        # client in pyoidc
+        self._client_extension = ClientExtension(client_authn_method=CLIENT_AUTHN_METHOD)
```
zamzterz, line 56: No need for this as a separate variable `oic.extension.client.Client` has allt the same methods as `oic.oic.Client` as far as I can see. So should be possible to just replace `self._client = ClientExtension(...)`.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -221,6 +225,56 @@ def userinfo_request(self, access_token):
...
+            request_args=args
+        )
+        logger.info('making token introspection request')
+        return self._client_extension.do_token_introspection(
+            request_args=token_introspection_request,
+            endpoint=self._client.introspection_endpoint)
+
+    def client_credentials_grant(self):
+        '''Public method to request access_token using client_credentials
+        (previously known as application in OpenApi2.0) flow. This is useful
```
zamzterz, line 256: Not sure about the relevance of the "OpenApi2.0" here, can it be removed?

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -221,6 +225,56 @@ def userinfo_request(self, access_token):
...
+        be used to access your APIs.
+        
+        How To Use?
+        -----------
+        >>>> auth = OIDCAuthentication({'default': provider_config},
+                                       access_token_required=True)
+        >>>> auth.init_app(app)
+        >>>> auth.clients['default'].client_credentials_grant()
+        '''
+        client_credentilas_payload = {
```
zamzterz, line 273: Tiny typo: "client_credenti**l**as_payload" -> "client_credentia**l**s_payload"

zamzterz (changes requested):
This looks amazing! 👏 

Huge kudos on the detailed docstrings.
I have some concerns about the mixing of responsibilities now within `oidc_auth`, I think it would be better to keep it separate from the existing user login in via browser that populates `flask.session`.

src/flask_pyoidc/flask_pyoidc.py
infohash: Initially, I implemented that design but the problem I faced with that was I was not able to test my endpoints with Swagger UI. I use falsk-restx to document APIs for flask app.

Within flask-restx, Swagger can be configured to use Oauth2 Authorization which sends token in the authorization header of the request. But because the endpoint is expecting cookie, it sends redirect response to Swagger instead of executing the view function.

Also, if you have a flask web app that has a developer feature to issue tokens to access your APIs, that won't work if APIs are made to work with either cookie only or token only. E.g. From Github developer settings, you can generate token to access Github APIs from your code that you normally do from Github web app.

Then I came up with the second approach for "why not expect both?" and go ahead with what is available. So if the user has valid cookie, access will be given, if not cookie, atleast expect token else if there's nothing, redirect the user.

I will push the demo app that I've created for this pull request once I'm done commenting it.

src/flask_pyoidc/pyoidc_facade.py
infohash: `construct_TokenInstrospectionRequest` and `do_token_introspection` are part of `oic.extension.client`. See https://github.com/OpenIDC/pyoidc/issues/638#issue-428418945

src/flask_pyoidc/flask_pyoidc.py
infohash: If the access_token is shared by the service to other service, would that be useful for the `view_func` to know about the token?

src/flask_pyoidc/flask_pyoidc.py
infohash: Earlier they were the same method but `oidc_decoraror` doesn't have to know about the token. It only needs to know if the authorization field in the request header is present or not. If it's there, `introspect_token` has the responsibility to have the token parsed from the request header and begin introspection. So I separated the responsibilities based on what is needed at what time.

src/flask_pyoidc/flask_pyoidc.py
zamzterz: I don't fully understand what issue you ran into with Swagger UI - it should support regular Authorization flow (though I do know there might be some issues with CORS).

But actually in terms of Github (and generally good API design) you shouldn't mix cookies and tokens as authorization.
As far as I know there's no public API that relies on cookies, instead you get a token (e.g. via Github developer settings) which you then use with defined REST APIs. Those same API endpoints do not accept cookies.

Typically tokens are used for authorization when there is no active user session, i.e. no cookies.
So I still think it's better to completely separate the concerns of auth via token or cookie.
That will help prevent accidental misuse.

src/flask_pyoidc/flask_pyoidc.py
zamzterz: What do you mean with 
> If the access_token is shared by the service to other service

In the case of token introspection here, I thought it would be for the app to be able to receive tokens for authorization.
If it doesn't suffice to know that a token was received, the `view_func` also needs to know *which* user the token was issued for (to be able to read any data related to that user).

src/flask_pyoidc/pyoidc_facade.py
zamzterz: You're right and that extension class also contains all the same methods as the base class.
So I'm thinking it should be possible to simply replace `self._client` with the extended class `ClientExtension` instead of adding `self._client_extension` as well.

src/flask_pyoidc/flask_pyoidc.py
infohash: In service to service architecture, Bob's service requests access token from the Identity Provider using Client Credentials grant. This access token can be used for two things:

1. Bob's service can access its own API using that token.
2. Bob's service can share that token to Alice's service so that her service can also access Bob's API. Bob's service can verify that token using token introspection. Only Bob can verify that token because it requires client_id and client_secret to make token introspection request.

So if the owner of that token is Bob, the view function of Bob's service will always see the userinfo of Bob which is not useful.

src/flask_pyoidc/pyoidc_facade.py
infohash: In https://github.com/OpenIDC/pyoidc/issues/638#issue-428418945 , they are discussing to move useful methods from the extension to the base class so we should not replace `_client` with `_client_extension`. Once the migration is complete, I'll remove the `_client_extension` object in future.

src/flask_pyoidc/flask_pyoidc.py
infohash: In order to separate it, I need to address one condition. What if there's a private API that is meant to be accessible for Alice who is accessing it through the web app in browser as well as for Bob who wants to access the API through curl or REST API client with access_token?

Should we allow to decorate the endpoint twice (one for OIDC & one for access_token)? I want to make it look seamless and with minimal changes to the user codebase. Something that can be turned on & off with an argument, per endpoint.

zamzterz: I don't expect that to be a very common use case, and the extra complexity is not worth it IMO when the user codebase work would be minimal. I imagine it actually being two separate endpoints that share the majority of code, something like:

```python
@app.route('/api/doWork')
@auth.token_auth(PROVIDER_NAME1)
def apiDoWork():
    authData = <use/extract token data>
    doWork(authData)

@app.route('/web/doWork')
@auth.oidc_auth(PROVIDER_NAME1)
def webDoWork():
    userSession = UserSession(flask.session)
    doWork(userSession)
```

src/flask_pyoidc/flask_pyoidc.py
zamzterz: With the service-to-service case: it's typically not the same client that introspects the token as the one that it was issued to.
What happens is usually:
1. Bob's service wants to access Alice's service, and obtains an access token from a common provider using Bob's client credentials.
2. Bob's service passes the token with the request to Alice's service.
3. Alice's service sends an introspection request with the token to the provider using Alice's client credentials.
4. Optional: the provide may restrict which client can introspect which tokens
5. Alice's service gets back the introspection response which identifies Bob's service as the "token owner" and may reject or accept Bob's service request.

The token might also be an access token issued for a user that has logged-in to Bob's service, i.e. issued after Authorization Code Flow (and not client credentials grant).
Hence it's necessary to expose the introspection information - it might be issued to a user and probably to another client than the service receiving it.

src/flask_pyoidc/pyoidc_facade.py
zamzterz: Now I see your point! 🙂 
However, that issue seems stale so I'd rather do it the other way around: only use `ClientExtension` for now and replace it back when/if the migration is completed in pyoidc. WDYT?

infohash: Not possible. These are the methods that are only available in `class Client` in `oic.oic` sub-package:

1. `store_registration_info` is fully implemented in base class and is partially implemented in extension class.
2. `construct_AuthorizationRequest` is only available in base class and this method is overridden method of its super class.
3. `verify_id_token` is only available in base class.
4. `do_user_info_request` is only available in base class.

src/flask_pyoidc/flask_pyoidc.py
infohash: Code duplication will deter developers to use this extension. In industrial projects, APIs grow quickly and if they happen to have this situation, they may have to use some other library if they don't want to duplicate the codebase of their API.

I've created another decorator `token_auth` that does token based authorization. But both `oidc_auth` & `token_auth` cannot be used together on the same view function. So it's either cookie based authentication or token based authorization which addresses your concern.

```
def token_auth(self, provider_name, scopes_required: list = None):
         '''Token based authorization.
 
         Parameters
         ----------
         provider_name : str
             Name of the provider registered with OIDCAuthorization.
         scopes_required : list, optional
             List of valid scopes associated with the endpoint.
 
         How To Use
         ----------
         >>> auth = OIDCAuthentication({'default': provider_config},
                                       access_token_required=True)
         >>> @app.route('/')
             @auth.token_auth(provider_name='default')
             def index():
                 ...
         >>> # You can also specify scopes required by your endpoint.
         >>> @auth.token_auth(provider_name='default',
                              scopes_required=['read', 'write'])
         '''
         def token_decorator(view_func):
 
             @functools.wraps(view_func)
             def wrapper(*args, **kwargs):
                 logger.debug('token executed')
                 client = self.clients[provider_name]
                 if not self._check_authorization_header(flask.request):
                     raise AccessDenied('missing authorization field')
                 token_introspection_result = self.introspect_token(
                     request=flask.request, client=client,
                     scopes=scopes_required)
                 if token_introspection_result:
                     logger.info('user has valid access token')
                     flask.session['token_introspection_info'] = token_introspection_result
                     return view_func(*args, **kwargs)
                 # Forbid access if the access token is invalid.
                 flask.abort(403)
             return wrapper
         return token_decorator
```
However, to address the problem with private API that can be made to accept both cookie and token, I've still kept the condition to check for token if cookie is not provided in `oidc_auth`. But this time it retrieves full information about the token and the user so overall it basically becomes decorator for authentication.

```
def oidc_auth(self, provider_name, accept_token: bool = None,
                   scopes_required: list = None):
         if provider_name not in self._provider_configurations:
             raise ValueError(
                 "Provider name '{}' not in configured providers: {}.".format(provider_name,
                                                                              self._provider_configurations.keys())
             )
         # If the user wants to enforce or disbale authorization on selective
         # endpoints, it can be enforced by providing boolean to accept_token
         # argument. If no boolean is explicitly provided, the value will be
         # taken from self.access_token_required which enforces authorization
         # globally and is False by default.
         if accept_token is None:
             accept_token = self.access_token_required
 
         def oidc_decorator(view_func):
             @functools.wraps(view_func)
             def wrapper(*args, **kwargs):
                 session = UserSession(flask.session, provider_name)
                 client = self.clients[session.current_provider]
 
                 if session.should_refresh(client.session_refresh_interval_seconds):
                     logger.debug('user auth will be refreshed "silently"')
                     return self._authenticate(client, interactive=False)
                 elif session.is_authenticated():
                     logger.debug('user is already authenticated')
                     return view_func(*args, **kwargs)
                 # If the user is not authenticated, check if the user has
                 # passed access token in the request header and make
                 # introspection request to the identity provider to verify the
                 # token.
                 elif accept_token and self._check_authorization_header(
                      flask.request):
                     token_introspection_result = self.introspect_token(
                         request=flask.request, client=client,
                         scopes=scopes_required)
                     if token_introspection_result:
                         logger.info('user has valid access token')
                         flask.session['token_introspection_info'] = token_introspection_result
                         return view_func(*args, **kwargs)
                     # Forbid access if the access token is invalid.
                     flask.abort(403)
                 # Else authenticate the user using Authorization Code Flow
                 else:
                     logger.debug('user not authenticated, start flow')
                     return self._authenticate(client)
 
             return wrapper
 
         return oidc_decorator
```
## What do you think?

---

### Current situation
flask_oidc extension is way outdated and [it's not complying with Oauth2.0 specs]( properly anymore so developers will likely adopt this extension if it takes little to no changes to their codebase. Based on the feedback that I've received from my team and community developers, these are the things they voted for:

1. A single decorator should be enough to protect APIs.
2. The integration with flask-pyoidc should require only minimal changes to the codebase.
3. It should be seamless as in it can turned on or off for selective endpoints by passing an argument.
4. A private API should be made accessible through both cookie based authentication and token based authorization.

src/flask_pyoidc/flask_pyoidc.py
infohash: Resolved. Now `introspect_token` returns entire token info containing information about the user as well.

src/flask_pyoidc/pyoidc_facade.py
zamzterz: Thanks for clarifying - I only skimmed the comparison between the two classes 👍

src/flask_pyoidc/flask_pyoidc.py
zamzterz: Sorry for my slow reply, holidays got in the way.
The `token_auth` method looks like what I was thinking with a separate decorator - thanks for sticking with it👍 .

I've realised yet one more reason for why I think fully splitting the 2 decorators though, i.e. not checking for any token i `oidc_auth` decorator, as it complicates the public API of this extension: the introspection response is stored under a separate key in the `flask.session` (`token_introspection_info`).
This means that if you're using the `oidc_auth` decorator with `accept_token=True` the user code still needs to do a slightly complicated check to see from which session key it should read the user data. This to me is a dealbreaker.

Also, I don't think the token introspection info should be stored in the `flask.session` at all.
If a token is used with a backend API (i.e. not related to any web frontend), I think it would be unexpected if it created a persistent session (returned as cookie).
It's a security risk, as it might lead to circumstances where a backend request is authenticated via cookies while it should only be authenticated via tokens.
Storing the introspection info in the request context (via `LocalProxy`) would be the way to go (see for example `current_user` from [flask-Login](https://github.com/maxcountryman/flask-login/blob/c760c0ef7ccc95d49b4693200245a4f2b148d41b/flask_login/utils.py#L25)).

infohash: I've committed new code to address your feedback. `token_introspection_result` will now be saved in `flask.g` application context which will be available for current request only. Once the user does `flask.g.pop('token_introspection_result', None)` to read token metadata, it is cleared from the application context. Even if the user doesn't pop it, it will be overwritten with new token_introspection_info value on new request.

I've given more thoughts on how to apply oidc as well as token authorization on private APIs and current implementation is the only way that works with no changes to the user codebase. I'm in touch with a dev team who has analytical web app and they have this situation. Based on the feedback I've received from devs, they don't want to duplicate their API endpoints.

Ideally, it makes sense to keep `oidc_auth` and `token_auth` completely separate but to address all possible scenarios, can you make current implementation an exception for this?

In future, a better solution and based on the feedback of other contributors, we can make amends to it. What do you think?

zamzterz: 1. I was thinking the same of putting the introspection info in `flask.g`, but per the [Flask documentation](https://flask.palletsprojects.com/en/2.0.x/appcontext/#storing-data) it should be reserved for only user-code. Hence I think it would be better to create a separate LocalProxy variable to guarantee it won't overwrite any user data (however unlikely that is).
2. I've experimented a bit and with your new addition of `token_auth` decorator it actually is possible to combine the decorators to support both token and cookie/web auth:
    ```python
    def token_and_cookie_auth(provider_name, scopes_required):
        def combined_decorator(view_func):
            @functools.wraps(view_func)
            def wrapper(*args, **kwargs):
                try:
                    return auth.token_auth(provider_name, scopes_required)(view_func)(*args, **kwargs)
                except HTTPException as e:
                    if e.code == 401: # TODO this relies on tiny change in `token_auth` to call `flask.abort(401)` instead of `raise AccessDenied`
                        # no token in the request, try web login flow instead
                        return auth.oidc_auth(provider_name)(view_func)(*args, **kwargs)

                    raise # will just pass-through the 403 response if there was a token but it was invalid

            return wrapper

        return combined_decorator

    @app.route('/protected')
    @token_and_cookie_auth('default', ['scope1'])
    def protected():
        # TODO check for introspection info or user session
    ```
  
   Would this option work with your requirements?
  
   The reason why I'm so hesitant to mix the two together as the first implementation is that once it has been added it will be very hard to make amends to it (it will be a major API change for this extension). So I want to explore all options where the existing `oidc_auth` decorator can be kept entirely outside this use case 🙂.

infohash: Earlier, I've used `_app_ctx_stack.top` as recommended by flask for extensions but our requirement is to allow user to obtain the information about the token so I used `flask.g` for this. If I use werkzeug's **LocalProxy**, will it be still possible for the user to obtain that information from application context?

Combining the decorators was a great solution. It precisely fits my requirement. I've created another decorator `access_control` based on your solution. The naming for this decorator I've taken from [this answer](https://security.stackexchange.com/a/109958). I've removed `_auth` suffix from its name because access_control is composed of authentication and authorization (too many auths here already). Naming suggestions are welcome.

```python
def access_control(self, provider_name: str,
                       scopes_required: list = None):
        '''This decorator serves dual purpose that is it can do both token
        based authorization and oidc based authentication. If your API needs
        to be accessible by either modes, use this decorator otherwise use
        either oidc_auth or token_auth.

        Parameters
        ----------
        provider_name : str
            Name of the provider registered with OIDCAuthorization.
        scopes_required : list, optional
            List of valid scopes associated with the endpoint.

        Raises
        ------
        HTTPException
            If the authorization is disabled.

        How To Use
        ----------
        >>> auth = OIDCAuthentication({'default': provider_config},
                                      disable_authorization=False)
        >>> @app.route('/')
            @auth.access_control(provider_name='default')
            def index():
                ...
        >>> # You can also specify scopes required by your endpoint.
        >>> @auth.access_control(provider_name='default',
                                 scopes_required=['read', 'write'])
        >>> # Additionally, you can disable token authorization entirely by
            # setting disable_authorization to True. It is helpful when you
            # want to deprecate developer access to your APIs.
        >>> auth = OIDCAuthentication({'default': provider_config},
                                      disable_authorization=True)
        '''
        def combined_decorator(view_func):

            fallback_to_oidc = self.oidc_auth(provider_name)(view_func)

            @functools.wraps(view_func)
            def wrapper(*args, **kwargs):

                try:
                    if self.disable_authorization:
                        logger.info('Authorization is disabled, this endpoint can use OIDC only')
                        flask.abort(401)
                    return self.token_auth(provider_name,
                                           scopes_required)(
                                               view_func)(*args, **kwargs)
                except HTTPException:
                    return fallback_to_oidc(*args, **kwargs)
                except AccessDenied:
                    return fallback_to_oidc(*args, **kwargs)
                except NotForMe:
                    return fallback_to_oidc(*args, **kwargs)

            return wrapper

        return combined_decorator
```

Now only `flask.g` is left to address once I receive your feedback. I'm also thinking to introduce token introspection for access_code saved in the cookie. If the user has cookie, we can have the access_code introspected along with checking the `last_authenticated` field.

zamzterz: About the `LocalProxy`: I've not had time to test and implement it myself, but looking at for example how flask-login does it I assume it's available to user code as well (I understand that as the entire point of [`current_user`](https://github.com/maxcountryman/flask-login/blob/c760c0ef7ccc95d49b4693200245a4f2b148d41b/flask_login/utils.py#L25)).

Looking at your implementation of `access_control` decorator it seems you have some additional custom requirements, namely the `disable_authorization` flag.
It got me thinking it would be better to leave this to the extension users to write this "combinated" decorator.
That would allow full control of how/when which access control mechanism (token or cookie) should be applied - for example to deprecate, or differences in handling of 401/403 when introspecting a token.
Your implementation falls back in all cases, but it could be argued that if a token was included in the request but invalid that no fallback should happen.
I'm not saying either case is more correct, rather that this extension should probably not be opinionated about this and leave it up to users to decide the proper access control.
All this is to say I'm leaning towards this `access_control` decorator being better placed as an example in the documentation rather than included in the public API. 🤔 

>  I'm also thinking to introduce token introspection for access_code saved in the cookie

Are you referring to the access_token stored in the session? (There's no "access_code".)
If so, what is the introspection needed for?
All user data should already be in the session in that case.

infohash: I've gone through werkzeug's **LocalProxy** and it has the same use case as flask `_app_ctx_stack.top`. They allow extensions to store internal data but it's not retrievable through **flask.g**. Our use case is to make `_token_introsoection_info` available to the user on application context. So if the extension wants to deliver that data to the user, flask.g is the way.

`disable_authorization` is not my own requirement but the reason of adding it is if your service is moving to version 2 APIs and you want to deprecate developer access to the version 1 APIs (maybe to encourage them to use version 2 API or maybe you only want the latest API version to have developer feature), then you can disable token based authorization for version 1 APIs with just one argument instead of making the user to revisit and modify the decorator. This is a lazy way but it's worth the time it saves to deprecate legacy API versions.

Asking users to implement hybrid functionality is not a good idea because users tend to write vulnerable code and security is hard. Besides, most developers prefer out of box support from extensions. I've received a feedback from a dev today and I'm told *"if the individual functionality is already offered by the extension, why can't it offer hybrid feature as well?"*.

I've committed new code which addresses when should the fallback to oidc happen. We should atleast offer the base hybrid functionality and if one isn't satisfied with the stock implementation, one can reimplement the hybrid decorator although it will be a rare usecase when one has to do it.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -491,11 +463,76 @@ def wrapper(*args, **kwargs):
                     logger.info('user has valid access token')
                     # Store token introspection info within the application
                     # context.
-                    g.token_introspection_info = token_introspection_result.to_dict()
+                    g._token_introspection_info = token_introspection_result.to_dict()
                     return view_func(*args, **kwargs)
                 # Forbid access if the access token is invalid.
-                flask.abort(403)
+                flask.abort(401)
```
zamzterz, line 469: This should still be a 403 - a token was present in the request but it was not valid.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -319,3 +334,205 @@ def valid_access_token(self, force_refresh=False):
...
+        '''
+        def token_decorator(view_func):
+
+            @functools.wraps(view_func)
+            def wrapper(*args, **kwargs):
+
+                client = self.clients[provider_name]
+                # Check for authorization field in the request header.
+                if not self._check_authorization_header(flask.request):
+                    raise AccessDenied('missing authorization field')
```
zamzterz, line 458: This should be `flask.abort(401)` to avoid having the users deal with this exception themselves.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -319,3 +334,205 @@ def valid_access_token(self, force_refresh=False):
...
+        logger.debug(result)
+        # Check if access_token is valid, active can be True or False
+        if not result.get('active'):
+            return False
+        # Check if client_id is in audience claim
+        if not client._client.client_id in result['aud']:
+            logger.info('Token is valid but required audience is missing')
+            # raises exception if client_id is not in audience, you can
+            # configure audience from Identity Provider
+            raise NotForMe('required audience is missing')
```
zamzterz, line 413: This should raise an exception that is propagated to user code.
I think it would suffice to log it (probably info level) and return `False` instead.

src/flask_pyoidc/flask_pyoidc.py
zamzterz: With the LocalProxy it won't be accessible via `flask.g`.
Instead this extension would create a variable that can be imported in user code and would contain the introspection data if it exists.

Fair enough about the hybrid decorator, you make a good case for including it. 👍 

I'm not convinced about the `disable_authorization` flag though - the deprecation of an auth method in an API would be an edge case and not really something this extension needs to be concerned about.
I think the cost - one more configuration option that needs to be maintained and documented - is too high for something that can be achieved with a simple find-and-replace operation.

I've added some comments about exceptions and 401/403 response code.
But after that I think this is close to being ready. 👏 
Could you please add some tests that cover the new functionality?

zamzterz (commented):
Added some thoughts about the handling of exceptions and 401/403.

src/flask_pyoidc/flask_pyoidc.py
infohash: If we don't raise exception, it will be difficult for the user to debug why authorization is failing despite the presence of a valid token. flask.abort 403 & 401 do not give necessary information about the failure. These failures will go undetected in monitoring dashboard that health checks microservices. Most users don't enable logging for packages so what went wrong won't show up on the console either.

src/flask_pyoidc/flask_pyoidc.py
infohash: I've added `LocalProxy` for token introspection data. It can be accessed by using `auth.current_identity` in the view function. It is available for current request only and is destroyed after every request just like `flask.g`.

src/flask_pyoidc/flask_pyoidc.py
zamzterz: My main reason for not raising an exception is that then will this require users of this extension to handle it themselves, instead of having it work right out of the box.
And in general it's a good thing to expose as little information as possible from the API itself to avoid it somehow being misused by attackers.

But a custom message might be helpful, however I think it should be even more generic.
Maybe something like: 'You do not have sufficient privileges for access.'

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -61,14 +61,14 @@ def __init__(self, provider_configurations, app=None,
         self.clients = None
         self._logout_view = None
         self._error_view = None
+        # current_identity proxy to obtain user info whose token was passed in
+        # the request. It is available until current request only and is
+        # destroyed between the request. The value is set by token_auth
+        # decorator.
+        self.current_identity = LocalProxy(lambda: getattr(_app_ctx_stack.top,
```
zamzterz, line 68: This looks great - really clean! 💯 

I was thinking about the name; how about `current_token_identity` or just `current_token` to clarify it will only contain data read from an introspected token?

infohash: I'm good with `current_token_identity`. It seems more descriptive.

src/flask_pyoidc/flask_pyoidc.py
infohash: Let's keep the default message then. It's good enough. I'll document possible failures in the documentation.

zamzterz:
Looking really really good! 
Are you able to add some tests covering this new functionality as well?

Other than tests some examples in [documentation](https://github.com/zamzterz/Flask-pyoidc/tree/main/docs) would be nice as part of this PR as well: for example how to configure the `OIDCAuthentication` object (needs `introspection_endpoint` in `ProviderMetadata`?) and how/when to use the new `token_auth` and `access_control` decorator.

infohash:
It will be ready to be merged by weekend.

zamzterz:
@infohash Just wanted to check-in, are you able to continue the work on this? 🙂

infohash:
Yes. I'll wrap up the work by this weekend. Sorry, I was caught up in my work which prevented me to complete the tasks here. Even I don't want to wait further to use these upcoming features. 🙂

zamzterz:
No problem! And no real rush on my end, just wanted to avoid leaving it hanging for too long.

infohash:
Test cases are done. Now I'll update the documentation accordingly. I'll also upload a full sample flask app with flask-restx demo that I have used to test these features.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -319,3 +333,203 @@ def valid_access_token(self, force_refresh=False):
...
+
+        Returns
+        -------
+        bool
+            True if access_token is valid else False.
+
+        Raises
+        ------
+        NotForMe
+            If access_token is invalid.
```
zamzterz, line 398: This no longer applies.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -319,3 +333,203 @@ def valid_access_token(self, force_refresh=False):
...
+
+        """
+        received_access_token = self._parse_access_token(request)
+        # send token introspection request
+        result = client._token_introspection_request(
+            access_token=received_access_token)
+        logger.debug(result)
+        # Check if access_token is valid, active can be True or False
+        if not result.get('active'):
+            return False
```
zamzterz, line 408: How about returning `None` instead (applies to all the early returns in this method)?
I think return type `typing.Optional[TokenIntrospectionResponse]` is more clear.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -319,3 +333,203 @@ def valid_access_token(self, force_refresh=False):
...
+                # Check for authorization field in the request header.
+                if not self._check_authorization_header(flask.request):
+                    logger.info('Request header has no authorization field')
+                    # Abort the request if authorization field is missing.
+                    flask.abort(401)
+                token_introspection_result = self.introspect_token(
+                    request=flask.request, client=client,
+                    scopes=scopes_required)
+                if token_introspection_result:
+                    logger.info('user has valid access token')
```
zamzterz, line 466: Nit: The token might be issued to a service instead of user, so more appropriate message would be something like "Request has valid access token".

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -221,6 +225,54 @@ def userinfo_request(self, access_token):
...
+        How To Use
+        ----------
+        >>> auth = OIDCAuthentication({'default': provider_config},
+                                      access_token_required=True)
+        >>> auth.init_app(app)
+        >>> auth.clients['default'].client_credentials_grant()
+        """
+        client_credentials_payload = {
+            'grant_type': 'client_credentials'
+            }
```
zamzterz, line 273: Formatting: this curly brace is incorrectly indendented.

tests/test_flask_pyoidc.py
```diff
@@ -22,7 +24,7 @@
 from .util import signed_id_token
 
 
-class TestOIDCAuthentication(object):
+class TestOIDCAuthentication(unittest.TestCase):
```
zamzterz, line 27: Why was this added?
~The test setup always uses `pytest` and no features from `unittest` so this should not be necessary.~
I saw it further down, with the assertions, but those should be changed to match the current style. 🙂

tests/test_flask_pyoidc.py
```diff
@@ -674,3 +676,221 @@ def test_should_return_None_if_token_refresh_request_fails(self):
...
+
+    def test_should_check_for_authorization_header(self):
+
+        authn = self.init_app()
+        with self.app.test_request_context('/'):
+            self.assertFalse(authn._check_authorization_header(flask.request))
+            flask.request.headers = {
+                'Authorization': 'Bearer access_token'
+            }
+            self.assertTrue(authn._check_authorization_header(flask.request))
```
zamzterz, line 688: This should be `assert authn._check_authorization_header(flask.request) == True` to match existing style (which doesn't use the unittest assert-helpers).
(This applies to all the added tests.)

tests/test_flask_pyoidc.py
```diff
@@ -674,3 +676,221 @@ def test_should_return_None_if_token_refresh_request_fails(self):
...
+            'introspection_endpoint': introspection_endpoint})
+        with self.app.test_request_context('/'):
+            flask.request.headers = {
+                'Authorization': 'Bearer access_token'
+            }
+            token_introspection_response = {
+                'active': True,
+                'aud': ['admin', 'user', self.CLIENT_ID],
+                'scope': 'read write delete',
+                'client_id': 'watchdogs'
```
zamzterz, line 769: This should be `self.CLIENT_ID` (applies to all cases of `client_id` below).

tests/test_flask_pyoidc.py
```diff
@@ -674,3 +676,221 @@ def test_should_return_None_if_token_refresh_request_fails(self):
...
+                flask.request, authn.clients[self.PROVIDER_NAME])
+            self.assertEqual(token_introspection_response,
+                             introspection_result.to_dict())
+
+    def test_token_auth_should_abort_request_if_authorization_missing(self):
+
+        authn = self.init_app()
+        view_mock = self.get_view_mock()
+        with self.app.test_request_context('/'):
+            self.assertRaises(HTTPException, authn.token_auth(
```
zamzterz, line 783: Would be good if this test checked the response code, which should be `401` when request is completely missing token in `Authorization` header.

tests/test_flask_pyoidc.py
```diff
@@ -674,3 +676,221 @@ def test_should_return_None_if_token_refresh_request_fails(self):
...
+            'scope': 'read write delete',
+            'client_id': 'watchdogs'
+        }
+        responses.add(responses.POST, introspection_endpoint,
+                      json=token_introspection_response)
+        with self.app.test_request_context('/'):
+            flask.request.headers = {
+                'Authorization': 'Bearer access_token'
+            }
+            self.assertRaises(HTTPException, authn.token_auth(
```
zamzterz, line 832: Same as above, would be good to check the response code is 403 in this case.

tests/test_flask_pyoidc.py
```diff
@@ -674,3 +676,221 @@ def test_should_return_None_if_token_refresh_request_fails(self):
...
+            'scope': 'read write delete',
+            'client_id': 'watchdogs'
+        }
+        responses.add(responses.POST, introspection_endpoint,
+                      json=token_introspection_response)
+        with self.app.test_request_context('/'):
+            flask.request.headers = {
+                'Authorization': 'Bearer access_token'
+            }
+            self.assertRaises(HTTPException, authn.access_control(
```
zamzterz, line 867: And again 🙂  - this should be a 403.

tests/test_pyoidc_facade.py
```diff
@@ -15,7 +17,7 @@
 REDIRECT_URI = 'https://rp.example.com/redirect_uri'
 
 
-class TestPyoidcFacade(object):
+class TestPyoidcFacade(unittest.TestCase):
```
zamzterz, line 20: Same as above, this should be replaced by using `assert`-statement instead.

zamzterz (changes requested):
Looking good, the added tests are really comprehensive 👍 !

Seems they don't run properly though on CI, but I think that's because of the change to use `unittest.TestCase` as base class. See specific review comments about that.

I spotted some small things, the main thing being the return value of `introspect_token`.

tests/test_flask_pyoidc.py
infohash: As `assert` evaluates to True or False, ` assert authn._check_authorization_header(flask.request)` is enough and complies with PEP8 convention.

zamzterz (approved):
This looks ready to me! 👍 

I'd be happy to merge it and add documentation as a separate PR, or are you already working on it?

infohash:
User documentation is also ready. #121

zamzterz:
Fantastic work @infohash, thank you for sticking with it! 🙇

PR #122, opened 2022-01-29
Update docs in Quickstart and Configuration.

zamzterz:
* Simplify code examples.
* Move "Client credentials flow" docs from Configuration to Quickstart.
* Update year in docs copyright notice.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@ sys.path.insert(0, os.path.abspath('..'))
 # -- Project information -----------------------------------------------------
 
 project = 'flask-pyoidc'
-copyright = '2019, Samuel Gulliksson'
+copyright = '2022, Samuel Gulliksson'
 author = 'Samuel Gulliksson'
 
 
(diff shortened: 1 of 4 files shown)


infohash:
```python
provider_config = ProviderConfiguration(client_metadata=client_metadata, [provider_configuration])

provider_config = ProviderConfiguration(client_registration_info=client_registration_info, [provider_configuration])
```

This looks confusing to the reader because it doesn't specify what keyword has to be provided to provide `provider_configuration`. The keyword can be either `issuer` or `provider_metadata`.

```python
provider_config = ProviderConfiguration(issuer='https://idp.example.com', [client configuration])
provider_config = ProviderConfiguration(provider_metadata=provider_metadata, [client configuration])
```

Same as above. `client_configuration` can be provided with `client_metadata` and `client_registration_info`.

It should be self descriptive so that the user can understand arguments just by looking at first line. Most users copy paste these lines so keyword arguments should be syntactically correct.

zamzterz:
I don't think putting together the two examples from your excerpts is too tricky.
Mainly I found, after looking at the rendered docs, that the configuration code blocks were long and even though well-commented, it still required reading through it to pick the correct configuration combination.

infohash:
Can  you add quickstart on PYPI? The page looks too empty.

zamzterz:
Great idea, fixed in #125 👍

PR #123, opened 2022-01-30
Added authenticated dynamic client registration

infohash:
Identity Providers support two ways how new clients can be registered through Client Registration Service.

1. **Authenticated requests** - Request to register new client must contain either Initial Access Token or Bearer Token.

2. **Anonymous requests** - Request to register new client doesn’t need to contain any token at all.

Features: 

1. Support for authenticated requests by using **Initial Access Token**.

2. Register multiple redirect URIs when the client is dynamically created.

---

Refer to [Client Registration](https://openid.net/specs/openid-connect-registration-1_0.html#ClientRegistration) and [Keycloak's Client Registration Service](https://github.com/keycloak/keycloak-documentation/blob/main/securing_apps/topics/client-registration.adoc#using-the-client-registration-service) for more info.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ dist/
 .coverage*
 coverage.xml
 .pytest_cache
-.scannerwork
 .venv/
 docs/_build
 docs/_static
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ dist/
 .coverage*
 coverage.xml
 .pytest_cache
+.scannerwork
 .venv/
 docs/_build
 docs/_static
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -185,6 +185,9 @@ class ProviderConfiguration:
         return self._client_metadata
 
     def register_client(self, redirect_uris, extra_parameters=None):
+
+        if not extra_parameters:
+            extra_parameters = {}
         if not self._client_metadata:
             if 'registration_endpoint' not in self._provider_metadata:
                 raise ValueError("Can't use dynamic client registration, provider metadata is missing "
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -200,7 +200,6 @@ class ProviderConfiguration:
             registration_response = client.register(
                 url=self._provider_metadata['registration_endpoint'],
                 **registration_request)
-            logger.debug(registration_response.to_dict())
             self._client_metadata = ClientMetadata(
                 **registration_response.to_dict())
             logger.debug('Received registration response: client_id=' + self._client_metadata['client_id'])
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -184,7 +184,7 @@ class ProviderConfiguration:
     def register_client(self, client: Client, redirect_uri: str):
 
         if not self._client_metadata:
-            if 'registration_endpoint' not in self._provider_metadata:
+            if not self._provider_metadata['registration_endpoint']:
                 raise ValueError("Can't use dynamic client registration, provider metadata is missing "
                                  "'registration_endpoint'.")
 
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -59,9 +59,15 @@ class OIDCData(collections.abc.MutableMapping):
 
 class ProviderMetadata(OIDCData):
 
-    def __init__(self, issuer=None, authorization_endpoint=None, jwks_uri=None,
-                 token_endpoint=None, userinfo_endpoint=None,
-                 introspection_endpoint=None, registration_endpoint=None, **kwargs):
+    def __init__(self,
+                 issuer=None,
+                 authorization_endpoint=None,
+                 jwks_uri=None,
+                 token_endpoint=None,
+                 userinfo_endpoint=None,
+                 introspection_endpoint=None,
+                 registration_endpoint=None,
+                 **kwargs):
         """OpenID Providers have metadata describing their configuration.
 
         Parameters
@@ -181,7 +187,7 @@ class ProviderConfiguration:
     def registered_client_metadata(self):
         return self._client_metadata
 
-    def register_client(self, client: Client, redirect_uri: str):
+    def register_client(self, client: Client):
 
         if not self._client_metadata:
             if not self._provider_metadata['registration_endpoint']:
@@ -189,22 +195,11 @@ class ProviderConfiguration:
                                  "'registration_endpoint'.")
 
             registration_request = self._client_registration_info.to_dict()
-            # Check if the user has passed list of redirect_uris in ClientRegistrationInfo.
-            if 'redirect_uris' not in registration_request:
-                # If not, create it.
-                registration_request['redirect_uris'] = [redirect_uri]
-            # if it's passed, check if it's type list.
-            elif not isinstance(registration_request['redirect_uris'], list):
-                raise ValueError('redirect_uris must be a list, not str.')
-            else:
-                # Add redirect_uri into the list of redirect_uris passed by
-                # the user in ClientRegistrationInfo.
-                registration_request['redirect_uris'].append(redirect_uri)
-            post_logout_redirect_uris = []
-            # Check if post_logout_redirect_uris exists.
-            if 'post_logout_redirect_uris' in registration_request:
-                post_logout_redirect_uris = registration_request['post_logout_redirect_uris']
-                # If it exists, add this into the list of redirect_uris.
+            # IdPs register all redirect URIs from redirect_uris parameter so
+            # post_logout_redirect_uris should be added to the redirect_uris
+            # parameter.
+            post_logout_redirect_uris = registration_request['post_logout_redirect_uris']
+            if post_logout_redirect_uris:
                 registration_request['redirect_uris'].extend(post_logout_redirect_uris)
             # Remove duplicate redirect_uris from the list.
             registration_request['redirect_uris'] = list(set(registration_request['redirect_uris']))
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -1,6 +1,8 @@
 import collections.abc
 import logging
 
+from oic.oic import Client
+from oic.utils.authn.client import CLIENT_AUTHN_METHOD
 import requests
 
 logger = logging.getLogger(__name__)
@@ -57,19 +59,35 @@ class OIDCData(collections.abc.MutableMapping):
 
 
 class ProviderMetadata(OIDCData):
-    def __init__(self, issuer=None, authorization_endpoint=None, jwks_uri=None, **kwargs):
-        """
-        Args:
-            issuer (str): OP Issuer Identifier.
-            authorization_endpoint (str): URL of the OP's Authorization Endpoint
-            jwks_uri (str): URL of the OP's JSON Web Key Set
-            kwargs (dict): key-value pairs corresponding to
-                `OpenID Provider Metadata`_
 
-        .. _OpenID Provider Metadata:
-            https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata
+    def __init__(self, issuer=None, authorization_endpoint=None, token_endpoint=None,
+                 userinfo_endpoint=None, jwks_uri=None, introspection_endpoint=None,
+                 registration_endpoint=None, **kwargs):
+        """OpenID Providers have metadata describing their configuration.
+
+        Parameters
+        ----------
+        issuer: str, Optional
+            OP Issuer Identifier.
+        authorization_endpoint: str, Optional
+            URL of the OP's OAuth 2.0 Authorization Endpoint.
+        token_endpoint: str, Optional
+            URL of the OP's OAuth 2.0 Token Endpoint.
+        userinfo_endpoint: str, Optional
+            URL of the OP's UserInfo Endpoint.
+        jwks_uri: str, Optional
+            URL of the OP's JSON Web Key Set [JWK] document.
+        introspection_endpoint: str, Optional
+            URL of the OP's token introspection endpoint.
+        registration_endpoint: str, Optional
+            URL of the OP's Dynamic Client Registration Endpoint.
+        **kwargs : dict, Optional
+            Extra arguments to [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata)
         """
-        super(ProviderMetadata, self).__init__(issuer=issuer, authorization_endpoint=authorization_endpoint, jwks_uri=jwks_uri, **kwargs)
+        super().__init__(issuer=issuer, authorization_endpoint=authorization_endpoint,
+                         token_endpoint=token_endpoint, userinfo_endpoint=userinfo_endpoint,
+                         jwks_uri=jwks_uri, introspection_endpoint=introspection_endpoint,
+                         registration_endpoint=registration_endpoint, **kwargs)
 
 
 class ClientRegistrationInfo(OIDCData):
@@ -149,6 +167,8 @@ class ProviderConfiguration:
 
         self.requests_session = requests_session or requests.Session()
 
+        self._client = Client(client_authn_method=CLIENT_AUTHN_METHOD)
+
     def ensure_provider_metadata(self):
         if not self._provider_metadata:
             resp = self.requests_session \
@@ -175,11 +195,12 @@ class ProviderConfiguration:
             if extra_parameters:
                 registration_request.update(extra_parameters)
 
-            resp = self.requests_session \
-                .post(self._provider_metadata['registration_endpoint'],
-                      json=registration_request,
-                      timeout=self.DEFAULT_REQUEST_TIMEOUT)
-            self._client_metadata = ClientMetadata(redirect_uris=redirect_uris, **resp.json())
+            registration_response = self._client.register(
+                url=self._provider_metadata['registration_endpoint'],
+                registration_token=registration_request.get('initial_access_token'),
+                **registration_request)
+            logger.debug(registration_response.to_dict())
+            self._client_metadata = ClientMetadata(**registration_response.to_dict())
             logger.debug('Received registration response: client_id=' + self._client_metadata['client_id'])
 
         return self._client_metadata
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -201,12 +201,6 @@ class ProviderConfiguration:
                 url=self._provider_metadata['registration_endpoint'],
                 **registration_request)
             logger.debug(registration_response.to_dict())
-            # Some IdPs may not return post_logout_redirect_uris to the
-            # response so explicitly patch the response to take
-            # post_logout_redirect_uris from the code. Even if it does exist
-            # in the response, it will be replaced.
-            registration_response['post_logout_redirect_uris'] = registration_request[
-                'post_logout_redirect_uris']
             self._client_metadata = ClientMetadata(
                 **registration_response.to_dict())
             logger.debug('Received registration response: client_id=' + self._client_metadata['client_id'])
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -59,9 +59,9 @@ class OIDCData(collections.abc.MutableMapping):
 
 class ProviderMetadata(OIDCData):
 
-    def __init__(self, issuer=None, authorization_endpoint=None, token_endpoint=None,
-                 userinfo_endpoint=None, jwks_uri=None, introspection_endpoint=None,
-                 registration_endpoint=None, **kwargs):
+    def __init__(self, issuer=None, authorization_endpoint=None, jwks_uri=None,
+                 token_endpoint=None, userinfo_endpoint=None,
+                 introspection_endpoint=None, registration_endpoint=None, **kwargs):
         """OpenID Providers have metadata describing their configuration.
 
         Parameters
@@ -70,12 +70,12 @@ class ProviderMetadata(OIDCData):
             OP Issuer Identifier.
         authorization_endpoint: str, Optional
             URL of the OP's OAuth 2.0 Authorization Endpoint.
+        jwks_uri: str, Optional
+            URL of the OP's JSON Web Key Set [JWK] document.
         token_endpoint: str, Optional
             URL of the OP's OAuth 2.0 Token Endpoint.
         userinfo_endpoint: str, Optional
             URL of the OP's UserInfo Endpoint.
-        jwks_uri: str, Optional
-            URL of the OP's JSON Web Key Set [JWK] document.
         introspection_endpoint: str, Optional
             URL of the OP's token introspection endpoint.
         registration_endpoint: str, Optional
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -195,14 +195,6 @@ class ProviderConfiguration:
                                  "'registration_endpoint'.")
 
             registration_request = self._client_registration_info.to_dict()
-            # IdPs register all redirect URIs from redirect_uris parameter so
-            # post_logout_redirect_uris should be added to the redirect_uris
-            # parameter.
-            post_logout_redirect_uris = registration_request['post_logout_redirect_uris']
-            if post_logout_redirect_uris:
-                registration_request['redirect_uris'].extend(post_logout_redirect_uris)
-            # Remove duplicate redirect_uris from the list.
-            registration_request['redirect_uris'] = list(set(registration_request['redirect_uris']))
 
             # Send request to register the client dynamically.
             registration_response = client.register(
@@ -214,7 +206,7 @@ class ProviderConfiguration:
             # ClientMetadata.
             self._client_metadata = ClientMetadata(
                 **registration_response.to_dict(),
-                post_logout_redirect_uris=post_logout_redirect_uris)
+                post_logout_redirect_uris=registration_request['post_logout_redirect_uris'])
             logger.debug('Received registration response: client_id=' + self._client_metadata['client_id'])
 
         return self._client_metadata
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -192,15 +192,13 @@ class ProviderConfiguration:
 
             registration_request = self._client_registration_info.to_dict()
             registration_request['redirect_uris'] = redirect_uris
-            if extra_parameters:
-                registration_request.update(extra_parameters)
 
             registration_response = self._client.register(
                 url=self._provider_metadata['registration_endpoint'],
-                registration_token=registration_request.get('initial_access_token'),
                 **registration_request)
             logger.debug(registration_response.to_dict())
-            self._client_metadata = ClientMetadata(**registration_response.to_dict())
+            self._client_metadata = ClientMetadata(**registration_response.to_dict(),
+                                                   **extra_parameters)
             logger.debug('Received registration response: client_id=' + self._client_metadata['client_id'])
 
         return self._client_metadata
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -2,7 +2,6 @@ import collections.abc
 import logging
 
 from oic.oic import Client
-from oic.utils.authn.client import CLIENT_AUTHN_METHOD
 import requests
 
 logger = logging.getLogger(__name__)
@@ -118,7 +117,7 @@ class ProviderConfiguration:
         session_refresh_interval_seconds (int): Number of seconds between updates of user data (tokens, user data, etc.)
             fetched from the provider. If `None` is specified, no silent updates should be made user data will be made.
         userinfo_endpoint_method (str): HTTP method ("GET" or "POST") to use when making the UserInfo Request. If
-            `None` is specifed, no UserInfo Request will be made.
+            `None` is specified, no UserInfo Request will be made.
     """
 
     DEFAULT_REQUEST_TIMEOUT = 5
@@ -167,8 +166,6 @@ class ProviderConfiguration:
 
         self.requests_session = requests_session or requests.Session()
 
-        self._client = Client(client_authn_method=CLIENT_AUTHN_METHOD)
-
     def ensure_provider_metadata(self):
         if not self._provider_metadata:
             resp = self.requests_session \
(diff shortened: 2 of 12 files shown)


infohash:
https://github.com/zamzterz/Flask-pyoidc/blob/40b1163e8055204d27930150ddccaa1da9bd487d/src/flask_pyoidc/pyoidc_facade.py#L72-L73

When the client is dynamically registered, the user may want to register multiple redirect_uris but in this line we are only registering one.

https://github.com/zamzterz/Flask-pyoidc/blob/40b1163e8055204d27930150ddccaa1da9bd487d/src/flask_pyoidc/provider_configuration.py#L174

Then this line adds only one redirect_uri in the registration request. This is also causing redirect to logout view to fail because `post_logout_redirect_uri` is not registered with Identity Provider's Valid Redirect URIs.

We need to make it this way so custom redirect_uris can be registered when the client is dynamically registered.

infohash:
Fixed https://github.com/zamzterz/Flask-pyoidc/pull/123#issuecomment-1025492365 and added custom redirect uris feature. 11 test cases are failing because of these changes, I'll address affected test cases.

infohash:
This feature can dynamically register all kinds of redirect URIs through Dynamic Client Registration so to register custom post logout rerdirect URIs in #124 dynamically, #123 is required.

infohash:
This will break lot of code if I split this. 1 & 2 are related to each other. I can write a review on my code to reference which are 1 & 2. Perhaps if you view from [Files changed](https://github.com/zamzterz/Flask-pyoidc/pull/123/files) section, it'll be much easier to review changes.

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
...
+            # Check if post_logout_redirect_uris exists.
+            if 'post_logout_redirect_uris' in registration_request:
+                post_logout_redirect_uris = registration_request['post_logout_redirect_uris']
+                # If it exists, add this into the list of redirect_uris.
+                registration_request['redirect_uris'].extend(post_logout_redirect_uris)
+            # Remove duplicate redirect_uris from the list.
+            registration_request['redirect_uris'] = list(set(registration_request['redirect_uris']))
+
+            # Send request to register the client dynamically.
+            registration_response = client.register(
```
infohash, line 213: This is authenticated dynamic client registration here. That's it. I'm simply using base library to make request. An optional `registration_token` argument is taken by `ClientRegistrationInfo` that takes `initial_access_token`. All arguments provoded by the user in `ClientRegistrationInfo` are in `registration_request` so I'm just unpacking this dictionary here (`**registration_request`) which contains `registration_token` argument if provided.

All other code changes that you see are for dynamically registering all redirect_uris provided by the user.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -269,10 +269,19 @@ def client_credentials_grant(self):
...
+        ::
+
+            auth.clients['default'].client_credentials_grant(
+                scope=['read', 'write'])
         """
         client_credentials_payload = {
             'grant_type': 'client_credentials'
         }
+        if scope:
+            client_credentials_payload['scope'] = ' '.join(scope)
```
infohash, line 284: A minor change, client credentials flow as per ouath specs can also take scope as an argument which takes string as a parameter delimited by spaces to separate multiple scopes. E.g. `scope='read write'`. Just to make it easier for the user, I'm taking it as a list and convert it into the string for the payload.

zamzterz:
While your comments help, I'd strongly prefer it to be split to follow the principle of "1 PR per change".
This PR is quite big (11 changed files) and only some of those actually relate to the title issue of this PR (include initial access token with client registration request).

It will help with the reviewing process (making it go faster) to have smaller chunks of changes to review.

infohash:
I'm not sure if 1 & 2 are separable. They may look different but they are part of the same flow that collects all redirect uris from the user code & then send them in the registration request. I implemented the 1st one but during testing I found it incomplete because it was registering only one redirect uri which lead to `invalid_redirect_uri` error. Then I worked on 2nd to complete the flow.

I should rather update PR description and title so it doesn't look like 2 different independent features.

zamzterz:
I might be misunderstanding something, but I don't see how 1 & 2 are related 🙂 
Yes, they are both part of client registration but
1. Adding support for including an access token in the request is only the addition of `registration_token` in `ClientRegistrationInfo` (which is then picked up by underlying pyoidc lib), while
2. Which redirect URIs are to be registered with the provider is independent of whether the request is authenticated (via token) or not.

Is it something specific for your use case/setup (which I gather is using Keycloak)?

infohash:
I think about new features and raise pull request. Keycloak is happen to be the one I'm using to test my features.

It's not worth the time to rework 1 & 2 separately because their test cases also have to be reverted and reworked. It broke existing test cases as well which I had to address them before writing new ones. It will duplicate my work to fix test cases again if I separate them.

I can draft a pull request for review purpose only. For merge, this pull request is the accurate one.

zamzterz:
Ok, leave it be for now. 🙂 

I'm having a pretty busy week, but I'll try to review it later this week. 👍

src/flask_pyoidc/provider_configuration.py
```diff
@@ -57,19 +58,35 @@ def to_dict(self):
...
-            https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata
+
+    def __init__(self, issuer=None, authorization_endpoint=None, jwks_uri=None,
+                 token_endpoint=None, userinfo_endpoint=None,
+                 introspection_endpoint=None, registration_endpoint=None, **kwargs):
+        """OpenID Providers have metadata describing their configuration.
+
+        Parameters
+        ----------
+        issuer: str, Optional
```
zamzterz, line 69: All optional types should be annotated `Optional[<type>]` to be consistent.

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
     def registered_client_metadata(self):
         return self._client_metadata
 
-    def register_client(self, redirect_uris, extra_parameters=None):
+    def register_client(self, client: Client, redirect_uri: str):
+
         if not self._client_metadata:
-            if 'registration_endpoint' not in self._provider_metadata:
+            if not self._provider_metadata['registration_endpoint']:
```
zamzterz, line 187: This will throw a `KeyError` if `registration_endpoint` is not set.
To safe-guard against empty value this check should be `if not self._provider_metadata.get('registration_endpoint')`

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
     def registered_client_metadata(self):
         return self._client_metadata
 
-    def register_client(self, redirect_uris, extra_parameters=None):
+    def register_client(self, client: Client, redirect_uri: str):
+
```
zamzterz, line 185: Extra empty line.

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
...
-                .post(self._provider_metadata['registration_endpoint'],
-                      json=registration_request,
-                      timeout=self.DEFAULT_REQUEST_TIMEOUT)
-            self._client_metadata = ClientMetadata(redirect_uris=redirect_uris, **resp.json())
+            # Check if the user has passed list of redirect_uris in ClientRegistrationInfo.
+            if 'redirect_uris' not in registration_request:
+                # If not, create it.
+                registration_request['redirect_uris'] = [redirect_uri]
+            # if it's passed, check if it's type list.
+            elif not isinstance(registration_request['redirect_uris'], list):
```
zamzterz, line 197: I don't think this check really belongs here - would be better in `ClientRegistrationInfo` if needed at all.

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
...
-            if extra_parameters:
-                registration_request.update(extra_parameters)
-
-            resp = self.requests_session \
-                .post(self._provider_metadata['registration_endpoint'],
-                      json=registration_request,
-                      timeout=self.DEFAULT_REQUEST_TIMEOUT)
-            self._client_metadata = ClientMetadata(redirect_uris=redirect_uris, **resp.json())
+            # Check if the user has passed list of redirect_uris in ClientRegistrationInfo.
+            if 'redirect_uris' not in registration_request:
```
zamzterz, line 193: For consistency, I think this should be handled outside `ProviderConfiguration` in the same way as `post_logout_redirect_uris`.
This layer shouldn't accept `redirect_uri` value that it potentially ignores.

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
...
+                # If not, create it.
+                registration_request['redirect_uris'] = [redirect_uri]
+            # if it's passed, check if it's type list.
+            elif not isinstance(registration_request['redirect_uris'], list):
+                raise ValueError('redirect_uris must be a list, not str.')
+            else:
+                # Add redirect_uri into the list of redirect_uris passed by
+                # the user in ClientRegistrationInfo.
+                registration_request['redirect_uris'].append(redirect_uri)
```
zamzterz, line 202: Not sure this is needed - if the user has specified redirect uris, it should be the complete list we pass to the provider without modification.

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
...
+            post_logout_redirect_uris = []
+            # Check if post_logout_redirect_uris exists.
+            if 'post_logout_redirect_uris' in registration_request:
+                post_logout_redirect_uris = registration_request['post_logout_redirect_uris']
+                # If it exists, add this into the list of redirect_uris.
+                registration_request['redirect_uris'].extend(post_logout_redirect_uris)
```
zamzterz, line 208: Why is it needed to duplicate `post_logout_redirect_uris` in `redirect_uri` list?
Could it be something provider specific?

src/flask_pyoidc/provider_configuration.py
zamzterz, line 213: This should not use `client.register` function, but keep the `self.requests_session` usage instead.
That might be configured differently, e.g. with timeouts, etc.

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +181,45 @@ def ensure_provider_metadata(self):
...
+            # Remove duplicate redirect_uris from the list.
+            registration_request['redirect_uris'] = list(set(registration_request['redirect_uris']))
+
+            # Send request to register the client dynamically.
+            registration_response = client.register(
+                url=self._provider_metadata['registration_endpoint'],
+                **registration_request)
+            logger.debug(registration_response.to_dict())
+            # post_logout_redirect_uris is not returned to the request so
```
zamzterz, line 217: This also seems like something provider specific (e.g. Keycloak), I don't think it should be needed for a provider properly supporting `post_logout_redirect_uris`.

src/flask_pyoidc/redirect_uri_config.py
```diff
@@ -60,4 +60,3 @@ def _parse_legacy_config(config):
         full_uri = scheme + '://' + redirect_domain + '/' + endpoint
 
         return full_uri, endpoint
-
```
zamzterz, line 63: Unrelated change.

tests/test_flask_pyoidc.py
```diff
@@ -162,19 +163,14 @@ def test_should_register_client_if_not_registered_before(self, post_logout_redir
...
-        responses.add(responses.POST, registration_endpoint, json={'client_id': 'client1', 'client_secret': 'secret1'})
+        responses.add(responses.POST, registration_endpoint, json={
+            'client_id': 'client1', 'client_secret': 'secret1',
+            'redirect_uris': ['https://client.example.com/redirect']})
         view_mock = self.get_view_mock()
         with self.app.test_request_context('/'):
             auth_redirect = authn.oidc_auth(self.PROVIDER_NAME)(view_mock)()
 
         self.assert_auth_redirect(auth_redirect)
-        registration_request = json.loads(responses.calls[0].request.body.decode('utf-8'))
```
zamzterz, line 171: Why was this removed?

tests/test_flask_pyoidc.py
```diff
@@ -890,3 +886,59 @@ def test_access_control_should_run_view_function_if_valid_token(self):
                 scopes_required=['read', 'write'])(view_mock)()
             assert view_mock.called
             assert flask._app_ctx_stack.top.current_token_identity == token_introspection_response
+
+    def test_get_url_for_logout_view(self):
```
zamzterz, line 890: How are this and the next test `test_get_url_for_logout_view_should_raise_build_error_if_mounted_under_custom_endpoint`?
Aren't they more related to #124?

tests/test_provider_configuration.py
```diff
@@ -57,41 +60,154 @@ def test_should_not_fetch_provider_metadata_if_given(self):
...
+        assert provider_config._client_metadata['client_name'] == 'Test Client'
+        assert provider_config._client_metadata['registration_client_uri'] == 'https://op.example.com/register/client1'
+        assert provider_config._client_metadata['registration_access_token'] == 'registration_access_token1'
+        assert provider_config._client_metadata['redirect_uris'] == [
+            'https://client.example.com/redirect', 'https://client.example.com/logout',
+            'https://client.example.com/swaggerui/oauth2-redirect.html']
+        assert provider_config._client_metadata[
+                   'post_logout_redirect_uris'] == ['https://client.example.com/logout']
+
+    def test_register_client_should_raise_value_error_if_type_redirect_uris_not_list(self):
```
zamzterz, line 162: As mentioned above, I don't think this behavior belongs in `ProviderConfiguration` 🤔

tests/test_provider_configuration.py
```diff
@@ -57,41 +60,154 @@ def test_should_not_fetch_provider_metadata_if_given(self):
...
         provider_config = ProviderConfiguration(provider_metadata=self.provider_metadata(),
                                                 client_registration_info=ClientRegistrationInfo())
+        assert provider_config._provider_metadata['registration_endpoint'] is None
         with pytest.raises(ValueError) as exc_info:
-            provider_config.register_client([])
-        assert 'registration_endpoint' in str(exc_info.value)
+            provider_config.register_client(None, None)
+
+    @responses.activate
+    def test_register_client_should_register_dynamic_client_if_initial_access_token_present(self):
```
zamzterz, line 98: This test should also assert the given `registration_token` was actually passed in the registration request:
Something like the following (which I've not tested it, so not 100% this is the way to access the headers from the captured request):
```python
assert responses.calls[0].request.headers['Authorization'] == 'Bearer initial_access_token'
```

tests/test_provider_configuration.py
```diff
@@ -57,41 +60,154 @@ def test_should_not_fetch_provider_metadata_if_given(self):
...
+    @responses.activate
+    def test_register_client_should_not_create_redirect_uris_key_if_already_passed(self):
+
+        registration_endpoint = self.PROVIDER_BASEURL + '/register'
+        client_registration_response = {
+            'client_id': 'client1',
+            'client_secret': 'secret1',
+            'client_name': 'Test Client',
+            'redirect_uris': ['https://client.example.com/redirect', 'https://client.example.com/logout',
+                              'https://client.example.com/swaggerui/oauth2-redirect.html'],
```
zamzterz, line 137: The `swaggerui` path element should be removed (applied to all places it occurs).

tests/test_provider_configuration.py
```diff
@@ -102,3 +218,9 @@ def test_copy_should_overwrite_existing_value(self):
         data = OIDCData(abc='xyz')
         copy_data = data.copy(qwe='rty', abc='123')
         assert copy_data == {'abc': '123', 'qwe': 'rty'}
+
+    def test_del_and_len(self):
```
zamzterz, line 222: I don't think this test is needed since those methods are directly delegated to the underlying storage.
(That's why there's only the case for `OIDCData.__str__` above - it's the only method except copy which has some custom behavior.)

tests/test_redirect_uri_config.py
```diff
@@ -31,6 +31,7 @@ def test_redirect_uri_config(self):
         redirect_uri_config = RedirectUriConfig.from_config(config)
         assert redirect_uri_config.full_uri == 'https://myexample.com:6000/callback'
         assert redirect_uri_config.endpoint == 'callback'
+        assert repr(redirect_uri_config) == f'({redirect_uri_config.full_uri}, {redirect_uri_config.endpoint})'
```
zamzterz, line 34: Unrelated to this PR and I'm not sure it's needed to test the string representation for this class (which isn't intended to be used for anything).

tests/test_pyoidc_facade.py
```diff
@@ -19,9 +17,9 @@
 
 class TestPyoidcFacade:
     PROVIDER_BASEURL = 'https://op.example.com'
-    PROVIDER_METADATA = ProviderMetadata(PROVIDER_BASEURL,
-                                         PROVIDER_BASEURL + '/auth',
-                                         PROVIDER_BASEURL + '/jwks')
+    PROVIDER_METADATA = ProviderMetadata(issuer=PROVIDER_BASEURL,
```
zamzterz, line 20: 👍

tests/test_pyoidc_facade.py
```diff
@@ -46,7 +44,15 @@ def test_is_registered(self):
...
-        responses.add(responses.POST, registration_endpoint, json=self.CLIENT_METADATA.to_dict())
+        client_registration_response = {
+            'client_id': 'client1',
+            'client_secret': 'secret1',
+            'client_name': 'Test Client',
+            'redirect_uris': ['https://rp.example.com/redirect_uri'],
+            'registration_client_uri': 'https://op.example.com/register/client1',
+            'registration_access_token': 'registration_access_token1'
+        }
+        responses.add(responses.POST, registration_endpoint, json=client_registration_response)
```
zamzterz, line 55: Why this change?

zamzterz (changes requested):
Great addition, thanks for taking the time to put this together! 🎉 

I think there's some unnecessary inconsistencies with how the redirect URIs are handled, both in implementation and behavior:
* specifying redirect_uris in client registration info should be passed unmodified to the provider
* `post_logout_redirect_uris` should not be mixed with `redirect_uris`
* determining whether redirect URIs have been specified should happen in `OIDCAuthentication._register_client`

Other than that it's mostly minor questions 🙂

.gitignore
```diff
@@ -8,6 +8,7 @@ dist/
 .coverage*
 coverage.xml
 .pytest_cache
+.scannerwork
```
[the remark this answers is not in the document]
infohash: sonarqube. I scan my code with this before committing.

src/flask_pyoidc/provider_configuration.py
infohash: Numpy docstring formatting rules are like this only.

src/flask_pyoidc/redirect_uri_config.py
infohash: PEP8 compliance: Leave a line at the end of the file.

tests/test_provider_configuration.py
infohash: It has increased the coverage.

tests/test_redirect_uri_config.py
infohash: To increase the test coverage. It's at 98.86% now.

src/flask_pyoidc/provider_configuration.py
infohash: It won't. `self._provider_metadata` is an object of `ProviderMetadata`, and not a dictionary. `self._provider_metadata['registration_endpoint']` is using `__getitem__` to fetch value of the `registration_endpoint` attribute of Provider Metadata class object which is default to `None`. So if the user doesn't provide registration endpoint URL, `self._provider_metadata['registration_endpoint']` returns None. If you use `self._provider_metadata.get('registration_endpoint')`, it will throw error that `_provider_metadata` has no attribute `get`.

src/flask_pyoidc/provider_configuration.py
infohash: Empty line after the function declaration is reserved for docstring.

src/flask_pyoidc/provider_configuration.py
infohash: `ClientRegistrationInfo` is an empty subclass of `OIDCData` so there was nowhere else to keep it.

src/flask_pyoidc/provider_configuration.py
infohash: `redirect_uris` is a standardized argument in [OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-connect-registration-1_0.html). To comply with this, we have to allow user to optionally pass this as an argument.

src/flask_pyoidc/provider_configuration.py
infohash: IdP only understands `redirect_uris` argument. So all post logout redirect uris are supposed to be in redirect_uris for IdP to register them. Otherwise it causes invalid redirect uri on logout.

src/flask_pyoidc/provider_configuration.py
infohash: I've given the reason why we should only use base package functions to interface with IdP in  #128.

src/flask_pyoidc/provider_configuration.py
infohash: `post_logout_redirect_uris` parameter is not a part of openid specs so no IdP supports it. To make it easy for the user, `ClientMetada` and pyoidc support `post_logout_redirect_uris` as an argument.

tests/test_flask_pyoidc.py
infohash: Now `client.register` method is registering the client which returns a response object. This line is no longer required.

tests/test_flask_pyoidc.py
infohash: This one is to increase test coverage. #124 came later.

tests/test_provider_configuration.py
infohash:
```python
assert provider_config._client_metadata['registration_access_token'] == 'registration_access_token1'
```
This has been asserted here.

tests/test_pyoidc_facade.py
infohash: To mock client registration response.

src/flask_pyoidc/provider_configuration.py
zamzterz: The `__getitem__` directly delegates to the underlying dictionary, so I see the following:
```python
>>> from flask_pyoidc.provider_configuration import ProviderMetadata
>>> a = ProviderMetadata()
>>> b = a['registration_endpoint']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "Flask-pyoidc/src/flask_pyoidc/provider_configuration.py", line 24, in __getitem__
    return self.store[key]
KeyError: 'registration_endpoint'
>>> b = a.get('registration_endpoint')
>>> print(b)
None
```

src/flask_pyoidc/provider_configuration.py
zamzterz: That could be extended to contain this check - I'm thinking in terms of class responsibility it would make more sense to make this check in that class instead of here (where it's also not checked until actually trying to use it).

src/flask_pyoidc/provider_configuration.py
zamzterz: For consistency it should use the existing setup (which might be improved at a later date).

src/flask_pyoidc/provider_configuration.py
zamzterz: Same as above, `post_logout_redirect_uris` is added to Client Registration data by a separate spec.
So they shouldn't be mixed.

src/flask_pyoidc/redirect_uri_config.py
zamzterz: Yes, great catch!
But since no other changes were made in this file in this PR it's not really related to this.
Minor formatting fixes like this would be ok for already changed files, but for bigger/other fixes a separate PR would be better (this might not be the only file in need of some cleaning 🙂 ).

tests/test_flask_pyoidc.py
zamzterz: This is checking what the registration request body was which should still be possible since pyoidc also uses `requests` library under the hood.
It does not depend on the response directly.

tests/test_flask_pyoidc.py
zamzterz: Increasing test coverage would be another issue better suited for a separate PR.

tests/test_provider_configuration.py
zamzterz: Those are not the same 🙂 :
* `assert responses.calls[0].request.headers['Authorization'] == 'Bearer initial_access_token'` <- checks the specified access token was included in the correct way in the registration request
* `assert provider_config._client_metadata['registration_access_token'] == 'registration_access_token1'` <- checks the parameter `registration_access_token` was returned in the registration response (also this parameter is not required to be returned by the IdP).

tests/test_provider_configuration.py
zamzterz: Same as above, it's unrelated to this PR and I really don't think it's necessary to test builtin dict-behavior.

tests/test_redirect_uri_config.py
zamzterz: While I appreciate your effort, striving for 100 % coverage is not really necessary. 🙂 
I think it's pushing against the limits of the trade-off of having more test code to maintain.

tests/test_pyoidc_facade.py
zamzterz: What was the difference needed for mocking the response like that instead of using the existing `CLIENT_METADATA`?

src/flask_pyoidc/provider_configuration.py
infohash: Keycloak also does. But it doesn't add URLs from post logout redirect uris in the list of valid redirect uris during client registration. That can be the behaviour with other IdPs also. In order to ensure that IdPs always register all post logout redirect uris, we have to add it in redirect_uris.

src/flask_pyoidc/provider_configuration.py
infohash: I did the same. It returns `None`. I have added all default arguments in `ProviderMetadata` class so `__getitem__` will always return `None` which is default value of the argument if the `registration_endpoint` is not provided by the user. You can test your code in my branch.

```python
class ProviderMetadata(OIDCData):

    def __init__(self,
                 issuer=None,
                 authorization_endpoint=None,
                 jwks_uri=None,
                 token_endpoint=None,
                 userinfo_endpoint=None,
                 introspection_endpoint=None,
                 registration_endpoint=None,
                 **kwargs):
        """OpenID Providers have metadata describing their configuration.

        Parameters
        ----------
        issuer: str, Optional
            OP Issuer Identifier.
        authorization_endpoint: str, Optional
            URL of the OP's OAuth 2.0 Authorization Endpoint.
        jwks_uri: str, Optional
            URL of the OP's JSON Web Key Set [JWK] document.
        token_endpoint: str, Optional
            URL of the OP's OAuth 2.0 Token Endpoint.
        userinfo_endpoint: str, Optional
            URL of the OP's UserInfo Endpoint.
        introspection_endpoint: str, Optional
            URL of the OP's token introspection endpoint.
        registration_endpoint: str, Optional
            URL of the OP's Dynamic Client Registration Endpoint.
        **kwargs : dict, Optional
            Extra arguments to [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata)
        """
        super().__init__(issuer=issuer, authorization_endpoint=authorization_endpoint,
                         token_endpoint=token_endpoint, userinfo_endpoint=userinfo_endpoint,
                         jwks_uri=jwks_uri, introspection_endpoint=introspection_endpoint,
                         registration_endpoint=registration_endpoint, **kwargs)
```

tests/test_pyoidc_facade.py
infohash: The response must have `redirect_uris`.

infohash:
Requested changes are done.

zamzterz (changes requested):
This is getting real close now!

The only main thing left is the mixing of `redirect_uris` and `post_logout_redirect_uris` (both in registration request and response).
I think this extension should opt for standards compliance so that special handling should not be included (and specifically Keycloak has an open issue for fixing this in their provider implementation).

src/flask_pyoidc/provider_configuration.py
infohash: PR to change rest of the raw requests is already complete. I'll open that one once this one is merged. It's not feasible for me to rework this because it's corresponding test cases takes unnecessary time to undo the change.

What we can do is once there are no more requested changes to do, I'll open that PR. Then you can merge both PR one after the another at the same time.

tests/test_pyoidc_facade.py
infohash: The response still has to be mocked to initialize `ClientMetadata` by using the parameters taken from the response. Then only it can be considered that `assert facade.is_registered() is True` is actually true.

src/flask_pyoidc/provider_configuration.py
zamzterz: All tests still pass if I simple revert this changed line to what it is in `main` branch: 
```python 
registration_response = client.register(
                url=self._provider_metadata['registration_endpoint'],
                **registration_request)
```

tests/test_pyoidc_facade.py
zamzterz: > Then only it can be considered that assert facade.is_registered() is True is actually true.

Not sure I'm following - the tests passed before this change and the actual values in the response does not matter.
What makes this change necessary for this tests?

src/flask_pyoidc/provider_configuration.py
```diff
@@ -164,22 +187,26 @@ def ensure_provider_metadata(self):
...
+            # explicitly provide this as a keyword argument to initialise
+            # ClientMetadata.
+            self._client_metadata = ClientMetadata(
+                **registration_response.to_dict(),
+                post_logout_redirect_uris=registration_request['post_logout_redirect_uris'])
```
zamzterz, line 209: This one as well now: `post_logout_redirect_uris=registration_request['post_logout_redirect_uris']` should not be needed.

infohash: Some IdPs may not return empty list if the user doesn't provide post_logout_redirect_uris  in `ClientRegistrationInfo` which will cause `ClientMetadata` to fail.

Here's the new code to ensure that `ClientMetada` takes `post_logout_redirect_uris` from the code instead of json response:

```python
# Send request to register the client dynamically.
registration_response = client.register(
    url=self._provider_metadata['registration_endpoint'],
    **registration_request)
logger.debug(registration_response.to_dict())
# Some IdPs may not return post_logout_redirect_uris to the
# response so explicitly patch the response to take
# post_logout_redirect_uris from the code. Even if it exists in
# the response, it will be replaced.
registration_response['post_logout_redirect_uris'] = registration_request[
    'post_logout_redirect_uris']
self._client_metadata = ClientMetadata(
    **registration_response.to_dict())
```

What do you think?

infohash:
Have you abandoned this project?

src/flask_pyoidc/provider_configuration.py
zamzterz: If the IdP doesn't return `post_logout_redirect_uris` or returns a subset of the URIs sent in the request it would be incorrect to overwrite it like this. It could indicate that provider either doesn't support `post_logout_redirect_uris` (and hence doesn't register any) or that they apply some policy to which URIs may be registered for a client (only accepting a subset of the ones that were sent).
Hence, the response from the provider should be taken as-is since we on this side can't be sure it wasn't intentional from the provider to not include the `post_logout_redirect_uris` in the response.

zamzterz:
> Have you abandoned this project?

Sorry for the long turnaround time on these PRs! 😞 
I have not abandoned this project, but it is more in a state of maintenance than active development.
And the time I have to spend on it is unfortunately very limited right now (mainly some small windows during weekends as you might have noticed).

src/flask_pyoidc/provider_configuration.py
infohash: Done 🙂

src/flask_pyoidc/provider_configuration.py
zamzterz: Still this one left, please change it back for simple merging of this before fixing all of these in the next PR.

zamzterz (commented):
Thanks for sticking with it! 👏

This looks ready to go with the minor fix of reverting the use of `requests_session` in `ProviderConfiguration.register_client`

src/flask_pyoidc/provider_configuration.py
infohash: This requires lot of rework to undo it, specially its test cases. You have to give an exception for this module. 😥

zamzterz: The changes in tests were really minor, but I'll make an exception this time.
I spotted a tiny security issue that I took the liberty to fix as well - so I'll merge this now.

Thanks so much again! 💯

zamzterz (approved):

infohash:
Until keycloak approves the feature to register post logout redirect uris https://github.com/keycloak/keycloak/discussions/9950, there's a workaround to register `post_logout_redirect_uris` with `ClientMetadata`. For anyone who ends up here with the issue:

```python
...
with app.app_context():
    auth._register_client(client)
client._provider_configuration._client_metadata[
    'post_logout_redirect_uris'] = ['https://client.example.com/logout', 'https://client.example.com/logout2']
client.register()
```

PR #124, opened 2022-02-01
Support for custom post logout redirect URIs

infohash:
Often times, a web app has 2 or more landing page when the user logs out. Some examples are:

1. If the user clicks on logout, it goes to the sign-in screen.
2. If the user clicks on "Delete My Account", it goes to the "Sorry to let you go", "Give us feedback", "Data Privacy", etc. screens.
3. If the user clicks on "Deactivate My Account", it can land on some informative page that tells the user about how to reactivate the account.

To address these cases, with this enhancement, the user can optionally provide an argument for the name of the `view function` which is resolved by `url_for(view_function_name)` to generate custom post logout redirect URI when the end session request is made.

https://github.com/zamzterz/Flask-pyoidc/blob/40b1163e8055204d27930150ddccaa1da9bd487d/src/flask_pyoidc/flask_pyoidc.py#L263

Currently, `_get_post_logout_redirect_uri` returns whatever post logout redirect URI is present at index zero.

https://github.com/zamzterz/Flask-pyoidc/blob/40b1163e8055204d27930150ddccaa1da9bd487d/src/flask_pyoidc/flask_pyoidc.py#L93

So even if the user has provided list of `post_logout_redirect_uris`  in `ClientMetadata` or in `ClientRegistrationInfo`, other URIs can't be used during the logout session.

`@auth.oidc_logout('name_of_the_view_function)` makes it possible to use other post logout redirect URI as well. This feature is backward compatible so just by decorating the view function with `@auth.oidc_logout` still works as it's.

> Note: If the user is not using Dynamic Client Registeration, all post logout redirect URIs have to be registered manually with the IdP's Valid Redirect URIs.

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -22,8 +22,6 @@ from urllib.parse import parse_qsl
 
 import flask
 import importlib_resources
-from flask import _app_ctx_stack, current_app
-from flask.helpers import url_for
 from oic import rndstr
 from oic.extension.message import TokenIntrospectionResponse
 from oic.oic import AuthorizationRequest
@@ -59,14 +57,14 @@ class OIDCAuthentication:
         self._provider_configurations = provider_configurations
 
         self.clients = None
-        self._logout_view = None
+        self._logout_view = []
         self._error_view = None
         # current_token_identity proxy to obtain user info whose token was
         # passed in the request. It is available until current request only and
         # is destroyed between the requests. The value is set by token_auth
         # decorator.
         self.current_token_identity = LocalProxy(lambda: getattr(
-            _app_ctx_stack.top, 'current_token_identity', None))
+            flask._app_ctx_stack.top, 'current_token_identity', None))
         self._redirect_uri_config = redirect_uri_config
 
         if app:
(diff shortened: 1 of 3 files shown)


zamzterz:
Thanks for identifying this issue - given your examples I understand the need for multiple post logout redirect URIs.

I think there's an even simpler way of achieving that, by supporting the `oidc_logout` decorator to be used multiple times.
I implemented a quick suggestion in #126, do you see any issues with that for your described use cases?

infohash:
@zamzterz Now we can address this one.

zamzterz:
> @zamzterz Now we can address this one.

Yes! I've actually just updated #126 with my proposal for solving the described issue.
Please have a look at that and let me know if that solves the use case.

infohash:
@zamzterz I have revised my approach for resolving custom post logout redirect uris. We can directly use `flask.request.url`.

```python
end_session_request = EndSessionRequest(
    id_token_hint=id_token_jwt,
    post_logout_redirect_uri=flask.request.url,
    state=flask.session['end_session_state'])
```

zamzterz:
Superseded by https://github.com/zamzterz/Flask-pyoidc/pull/126 that has now been merged! 🎉

PR #126, opened 2022-02-07
Support multiple logout views (multiple use of 'oidc_logout' decorator).

zamzterz:
~~*Note: Not tested yet*~~

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -59,7 +59,7 @@ class OIDCAuthentication:
         self._provider_configurations = provider_configurations
 
         self.clients = None
-        self._logout_view = None
+        self._logout_views = []
         self._error_view = None
         # current_token_identity proxy to obtain user info whose token was
         # passed in the request. It is available until current request only and
@@ -88,42 +88,22 @@ class OIDCAuthentication:
             for (name, configuration) in self._provider_configurations.items()
         }
 
-    def _get_post_logout_redirect_uri(self, client):
-        if client.post_logout_redirect_uris:
-            return client.post_logout_redirect_uris[0]
-        return self._get_url_for_logout_view()
-
-    def _get_url_for_logout_view(self):
-        if not self._logout_view:
-            return None
-
+    def _get_urls_for_logout_views(self):
         try:
-            return url_for(self._logout_view.__name__, _external=True)
+            return [url_for(view.__name__, _external=True) for view in self._logout_views]
         except BuildError:
             logger.error('could not build url for logout view, it might be mounted under a custom endpoint')
             raise
 
     def _register_client(self, client):
-        def default_post_logout_redirect_uris():
-            url_for_logout_view = self._get_url_for_logout_view()
-            if url_for_logout_view:
-                return [url_for_logout_view]
-            return []
-
-        # Check if the user has passed list of redirect_uris in ClientRegistrationInfo.
         if not client._provider_configuration._client_registration_info.get('redirect_uris'):
-            # If not, create it.
             client._provider_configuration._client_registration_info[
                 'redirect_uris'] = [self._redirect_uri_config.full_uri]
-        # Check if the user has passed post_logout_redirect_uris in ClientRegistrationInfo.
         post_logout_redirect_uris = client._provider_configuration._client_registration_info.get(
             'post_logout_redirect_uris')
         if not post_logout_redirect_uris:
-            # If not passed, try to resolve it by using logout view function.
-            _default_post_logout_redirect_uris = default_post_logout_redirect_uris()
-            # Set this as an attribute of ClientRegistrationInfo.
             client._provider_configuration._client_registration_info[
-                'post_logout_redirect_uris'] = _default_post_logout_redirect_uris
+                'post_logout_redirect_uris'] = self._get_urls_for_logout_views()
         logger.debug(
             f'''registering with post_logout_redirect_uris = {
                 client._provider_configuration._client_registration_info[
@@ -254,7 +234,7 @@ class OIDCAuthentication:
 
         return oidc_decorator
 
-    def _logout(self):
+    def _logout(self, view_func):
         logger.debug('user logout')
         try:
             session = UserSession(flask.session)
@@ -269,8 +249,9 @@ class OIDCAuthentication:
         if client.provider_end_session_endpoint:
             flask.session['end_session_state'] = rndstr()
 
+            post_logout_uri = url_for(view_func.__name__, _external=True)
             end_session_request = EndSessionRequest(id_token_hint=id_token_jwt,
-                                                    post_logout_redirect_uri=self._get_post_logout_redirect_uri(client),
+                                                    post_logout_redirect_uri=post_logout_uri,
                                                     state=flask.session['end_session_state'])
 
             logger.debug('send endsession request: %s', end_session_request.to_json())
@@ -279,7 +260,7 @@ class OIDCAuthentication:
         return None
 
     def oidc_logout(self, view_func):
-        self._logout_view = view_func
+        self._logout_views.append(view_func)
 
         @functools.wraps(view_func)
         def wrapper(*args, **kwargs):
@@ -289,7 +270,7 @@ class OIDCAuthentication:
                     logger.error("Got unexpected state '%s' after logout redirect.", flask.request.args['state'])
                 return view_func(*args, **kwargs)
 
-            redirect_to_provider = self._logout()
+            redirect_to_provider = self._logout(view_func)
             if redirect_to_provider:
                 return redirect_to_provider
 
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -90,10 +90,17 @@ class OIDCAuthentication:
 
     def _get_urls_for_logout_views(self):
         try:
-            return [url_for(view.__name__, _external=True) for view in self._logout_views]
+            return [self._get_url_for_view_function(view) for view in self._logout_views]
         except BuildError:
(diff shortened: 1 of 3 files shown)


infohash:
I left a review highlighting the problem.

zamzterz:
@infohash I don't see any review? Was it not published because this PR was still in draft stage?

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -269,8 +249,9 @@ def _logout(self):
         if client.provider_end_session_endpoint:
             flask.session['end_session_state'] = rndstr()
 
+            post_logout_uri = url_for(view_func.__name__, _external=True)
```
infohash, line 252: This will fail for logout view functions that are mounted under blueprints. Name of view functions under a blueprint are prefixed by the name of the associated blueprint. In large-scaled flask apps, this is very common. A user should have an option available to provide view function name that can be resolved by `url_for`.

zamzterz: You're right - I opted to leave this behavior for the draft as that's how it was before.

I've just pushed a change now to make it more general, with tests that show it works regardless of which endpoint the view function is mounted under.
The benefit as I see it with this PR is that it will automatically handle the most common cases (including setting the correct `post_logout_redirect_uris` in the registration request) without any changes from the user side.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -88,42 +88,29 @@ def init_app(self, app):
...
 
-    def _register_client(self, client):
-        def default_post_logout_redirect_uris():
-            url_for_logout_view = self._get_url_for_logout_view()
-            if url_for_logout_view:
-                return [url_for_logout_view]
-            return []
+    def _get_url_for_view_function(self, view_func):
+        for endpoint, f in current_app.view_functions.items():
+            if f == view_func:
```
infohash, line 100: This if condition will never be true if you are using flask extensions that wrap view functions to generate OpenAPI documentation (Swagger). For example if you use `flask-restx`, the view function object becomes `<function Logout.get at 0x000001DF5DB27760>` and is registered with flask app as:

```python
>>> app.view_functions
{'apiv2.logout_logout': <function View.as_view.<locals>.view at 0x000001DF5DB40670>
```

In an enterprise environment, OpenAPI compliance is a standard so all view functions are likely to be wrapped with extensions that automatically generate swagger documentation for APIs. We must not have conflicts with the use of other flask-extensions.

zamzterz: That is due to the ordering of the decorators: `@auth.oidc_logout` needs to be the second outermost decorator, right after the Flask `@app.route(...)`:
```python
@app.route('/logout')
@auth.oidc_logout
@access_log
@....
def logout():
    return "You've been successfully logged out!"
```

infohash: I did it in #124. I'm using `flask.request.url` instead.

zamzterz: I think it would be good if we aligned on a single PR to not duplicate the efforts 🙂 .

I checked #124 and it has some unrelated changes now (e.g. the imports), and this one is ready to go, so I'd like to merge this one.
Is there anything more that should be added to this one from #124?

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -88,42 +88,22 @@ def init_app(self, app):
             for (name, configuration) in self._provider_configurations.items()
         }
 
-    def _get_post_logout_redirect_uri(self, client):
-        if client.post_logout_redirect_uris:
```
infohash, line 92: As this method is no longer required, add a test case to cover `post_logout_redirect_uris` property of `PyoidcFacade`. Take it from my PR.

```python
@property
def post_logout_redirect_uris(self):
    return self._client.registration_response.get('post_logout_redirect_uris')

def test_property_post_logout_redirect_uris(self):
    post_logout_redirect_uris = ['https://client.example.com/logout']
    client_metadata = self.CLIENT_METADATA.copy(
        post_logout_redirect_uris=post_logout_redirect_uris)
    facade = PyoidcFacade(ProviderConfiguration(provider_metadata=self.PROVIDER_METADATA,
        client_metadata=client_metadata),
                              REDIRECT_URI)
    assert facade.post_logout_redirect_uris == post_logout_redirect_uris
```

infohash (changes requested):

src/flask_pyoidc/flask_pyoidc.py
infohash: Our approach is exactly same 😂. I have left some reviews to address the changes.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -88,42 +88,22 @@ def init_app(self, app):
...
-        return self._get_url_for_logout_view()
-
-    def _get_url_for_logout_view(self):
-        if not self._logout_view:
-            return None
-
+    def _get_urls_for_logout_views(self):
         try:
-            return url_for(self._logout_view.__name__, _external=True)
+            return [url_for(view.__name__, _external=True) for view in self._logout_views]
```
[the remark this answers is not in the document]
zamzterz: I'd say this is an expected error, and the given error message points to the solution as well: with `SERVER_NAME` in the Flask config it works as expected:
```python
app.config.update({'SERVER_NAME': 'foobar', ...})
auth.init_app(app)
client = auth.clients[PROVIDER_NAME1]
with app.app_context():
    auth._register_client(client)
```

This is standard Flask behavior, that it needs to either have a request context or know the `SERVER_NAME` to be able to figure out the URL for and endpoint.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -88,42 +88,22 @@ def init_app(self, app):
...
         }
 
-    def _get_post_logout_redirect_uri(self, client):
-        if client.post_logout_redirect_uris:
-            return client.post_logout_redirect_uris[0]
-        return self._get_url_for_logout_view()
-
-    def _get_url_for_logout_view(self):
-        if not self._logout_view:
-            return None
```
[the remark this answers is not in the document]
zamzterz: This will already return an empty list if `self._logout_views` is empty.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -88,42 +88,22 @@ def init_app(self, app):
...
             client._provider_configuration._client_registration_info[
                 'redirect_uris'] = [self._redirect_uri_config.full_uri]
-        # Check if the user has passed post_logout_redirect_uris in ClientRegistrationInfo.
         post_logout_redirect_uris = client._provider_configuration._client_registration_info.get(
             'post_logout_redirect_uris')
         if not post_logout_redirect_uris:
-            # If not passed, try to resolve it by using logout view function.
-            _default_post_logout_redirect_uris = default_post_logout_redirect_uris()
-            # Set this as an attribute of ClientRegistrationInfo.
             client._provider_configuration._client_registration_info[
```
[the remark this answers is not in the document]
zamzterz: I found the comments in this case redundant when looking over them - they don't explain any additional info that is not already in the code itself.
Or am I missing something?

tests/test_flask_pyoidc.py
```diff
@@ -972,25 +1027,3 @@ def test_access_control_should_run_view_function_if_valid_token(self):
...
-
-        authn = self.init_app()
-        assert authn._get_url_for_logout_view() is None
-        logout_view_mock = self.get_view_mock()
-        authn.oidc_logout(logout_view_mock)
-        self.app.add_url_rule('/logout', view_func=logout_view_mock)
-        with self.app.test_request_context('/'):
-            assert authn._get_url_for_logout_view() == f'http://{self.CLIENT_DOMAIN}/logout'
-
-    def test_get_url_for_logout_view_should_raise_build_error_if_mounted_under_custom_endpoint(self):
```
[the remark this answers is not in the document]
zamzterz: The method `_get_url_for_logout_view` no longer exists so this test didn't make sense after these changes.

src/flask_pyoidc/flask_pyoidc.py
zamzterz: Fixed! 👍

src/flask_pyoidc/flask_pyoidc.py
zamzterz: Hehe! :)

Thanks a ton for the detailed review, I've addressed the comments! 🙇

tests/test_flask_pyoidc.py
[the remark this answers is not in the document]
infohash: It does.

https://github.com/zamzterz/Flask-pyoidc/blob/9d08533e97315dd97f89730d3e1bffa6bd9bd74f/src/flask_pyoidc/flask_pyoidc.py#L91-L96

infohash:
Test coverage is at 99.20%. We can easily make it 100% now. Striving for 100% is good when the code base is not large yet. This will benefit us in future PRs. You can set test coverage threshold to 98% now.

How do you bump versions? Should we keep our version scheme consistent with Python? Like keeping it `3.10.0` - `3.10.x` until Python 3.11 version is released?

zamzterz:
@infohash I don't think 100% coverage is necessary. And actually I'd like to refactor the tests sometime™️  for less test code to maintain - that's higher priority than increasing already high coverage. 🙂 

As for the versioning: this extension is using semantic versioning, so not tied to Python versioning at all.
Next version would be 3.11.0, do you think it should be released now?
Otherwise I was planning on holding off since you mentioned you were working on the "Client/Http session" refactor already.

infohash:
@zamzterz I have delegated all requests to underlying pyoidc library but test cases are failing because `oic.oic.Client` methods are not returning `id_token_jwt`. I'll have to raise PR there to support this feature.

PR #127, opened 2022-02-09
Added scope and extra keyword arguments that can be optionally passed to client crerdentials flow

infohash:
To comply with 4.4.2.  Access Token Request of Client Credentials Grant in  RFC 6749.

### DIFF
--- a/src/flask_pyoidc/pyoidc_facade.py
+++ b/src/flask_pyoidc/pyoidc_facade.py
@@ -277,7 +277,7 @@ class PyoidcFacade:
             auth.init_app(app)
             auth.clients['default'].client_credentials_grant()
 
-        Optionally, you can also specify scope for the access token.
+        Optionally, you can specify scopes for the access token.
 
         ::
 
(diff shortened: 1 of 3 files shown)


zamzterz (approved):
Great - thanks! 👍

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -261,6 +261,13 @@ def client_credentials_grant(self):
         On API call, token introspection will ensure that only valid token can
         be used to access your APIs.
 
+        Parameters
+        ----------
+        scope: list, optional
```
zamzterz, line 266: Type should be `Optional[list]`.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -261,6 +261,13 @@ def client_credentials_grant(self):
...
+            List of scopes to be requested.
+        **kwargs : dict, optional
```
zamzterz, line 268: Type should be `Optional[dict]`.

src/flask_pyoidc/pyoidc_facade.py
infohash: Numpy docstring format rules say to annotate docstring parameters like this only.

zamzterz: Right, I've just reviewed those and the Google Python style guide and seems it's changed compared to the current format in this code base.
I'm thinking about dropping the types in docs and moving fully to proper type annotations.

PR #128, opened 2022-02-12
Use ProviderConfiguration.requests_session for all requests.

zamzterz:
Also switch to Basic auth for token introspection request.

### DIFF
--- a/src/flask_pyoidc/pyoidc_facade.py
+++ b/src/flask_pyoidc/pyoidc_facade.py
@@ -3,8 +3,9 @@ import json
 import logging
 
 from oic.extension.client import Client as ClientExtension
+from oic.extension.message import TokenIntrospectionResponse
 from oic.oic import Client, RegistrationResponse, AuthorizationResponse, \
-    AccessTokenResponse, TokenErrorResponse, AuthorizationErrorResponse
+    AccessTokenResponse, TokenErrorResponse, AuthorizationErrorResponse, OpenIDSchema
 from oic.oic.message import ProviderConfigurationResponse
 from oic.utils.authn.client import CLIENT_AUTHN_METHOD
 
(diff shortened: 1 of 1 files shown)


src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -220,10 +221,12 @@ def userinfo_request(self, access_token):
             return None
 
         logger.debug('making userinfo request')
-        userinfo_response = self._client.do_user_info_request(method=http_method, token=access_token)
```
infohash, line 223: We shouldn't interface with IdP directly. This extension is a wrapper around pyoidc which complies with openid specs and is a [Certified OpenID Connect Implementations](https://openid.net/developers/certified/). Let base package to handle compliance with openid specs and interfacing with IdP. Pyoidc should be the one making these requests so that if anything changes in future versions of openid specs, they can adjust request and response with IdP accordingly.

Apart from the parameters that we are taking from the user, there can be additional parameters that maybe default or aren't important enough to be provided by the user unless the user explicitly wants to override the behaviour. Pyoidc ensures maintainability of these things and will also address changes in the standard in future versions. If you check [`do_user_info_request`](https://github.com/OpenIDC/pyoidc/blob/bd510e49ebd6d8816e3ee94e62a3e86e907a0f8d/src/oic/oic/__init__.py#L934), it can take additional keyword arguments that the user can provide for specific usecases. But now because of these raw requests to IdP, there's no such customization possible and responsibility of custom arguments falls on us to provide to the user.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -238,18 +241,18 @@ def _token_introspection_request(self, access_token: str):
...
-            request_args=args
-        )
+        auth_header = _ClientAuthentication(self._client.client_id, self._client.client_secret)('client_secret_basic',
+                                                                                                request)
         logger.info('making token introspection request')
-        return self._client_extension.do_token_introspection(
-            request_args=token_introspection_request,
-            endpoint=self._client.introspection_endpoint)
+        response = self._provider_configuration.requests_session \
+            .post(self._client.introspection_endpoint, data=request, headers=auth_header) \
```
infohash, line 252: Same as above reason. If we don't let pyoidc to interface with IdP, it becomes our responsibility to comply with openid specs. In future PRs, I would have rather replace existing raw requests to the IdP with the functions implemented in pyoidc. I prefer reusing and wrapping the base code over rewriting the functionality.

src/flask_pyoidc/pyoidc_facade.py
zamzterz: You make a good point, but this is a trade-off I've decided to make due to a real-word use case where more control of the network requests was needed (to be able to configure retries and timeouts).
I don't expect the requests/response to change - it's been a stable spec for many years.

However, this PR was only a quick fix to get consistency (there shouldn't be two different ways in how the HTTP requests are made).
A future option could be to subclass Client and re-implement `http_request` instead to use a custom `Requests.Session`.
It would avoid the custom request/response handling entirely while still allow the necessary control of the networking.

PR #132, opened 2022-03-27
Delegated all requests to pyoidc library

infohash:
All requests for IdP are now entrusted to base library.

# Changelogs

## Delegations
1. `flask_pyoidc.pyoidc_facade.PyoidcFacade.parse_authentication_response` -> `oic.oic.Client.parse_response`
2. `flask_pyoidc.pyoidc_facade.PyoidcFacade.refresh_token` -> `oic.oic.Client.do_access_token_refresh`
3. `flask_pyoidc.pyoidc_facade.PyoidcFacade._token_request` -> `oic.oic.Client.do_access_token_request`
4. `flask_pyoidc.pyoidc_facade.PyoidcFacade.userinfo_request` ->  `oic.oic.Client.do_user_info_request`
5. `flask_pyoidc.pyoidc_facade.PyoidcFacade._token_introspection_request` -> `oic.oic.ClientExtension.do_token_introspection`
6. `flask_pyoidc.pyoidc_facade.PyoidcFacade.client_credentials_grant` -> `oic.oauth2.Client.do_access_token_request`
7. `flask_pyoidc.pyoidc_facade.provider_configuration` -> `oic.oic.Client.provider_config`

## Additions
1. Added message_factory module for Client Credentials Message Factory.
2. Added `oic.utils.settings.ClientSettings` for `requests.Session` persistence. (Depends on OpenIDC/pyoidc#806)

## Tests

Their test cases are modified with respect to https://github.com/OpenIDC/pyoidc/pull/805 which adds support for including `id_token_jwt` in token response.

---

Drafting this PR until the latest version of pyoidc brings `id_token_jwt` support.

### DIFF
--- a/src/flask_pyoidc/message_factory.py
+++ b/src/flask_pyoidc/message_factory.py
@@ -1,0 +1,6 @@
+from oic.oauth2.message import AccessTokenResponse, CCAccessTokenRequest, MessageTuple, OauthMessageFactory
+
+
+class CCMessageFactory(OauthMessageFactory):
+    """Client Credential Request Factory."""
+    token_endpoint = MessageTuple(CCAccessTokenRequest, AccessTokenResponse)
--- a/src/flask_pyoidc/pyoidc_facade.py
+++ b/src/flask_pyoidc/pyoidc_facade.py
@@ -27,16 +27,16 @@ class PyoidcFacade:
         """
         self._provider_configuration = provider_configuration
         self._client = Client(client_authn_method=CLIENT_AUTHN_METHOD,
-                              settings=provider_configuration.requests_session)
+                              settings=provider_configuration.client_settings)
         # Token Introspection is implemented under extension sub-package of
         # the client in pyoidc.
         self._client_extension = ClientExtension(client_authn_method=CLIENT_AUTHN_METHOD,
-                                                 settings=provider_configuration.requests_session)
+                                                 settings=provider_configuration.client_settings)
         # Client Credentials Flow is implemented under oauth2 sub-package of
         # the client in pyoidc.
         self._oauth2_client = Oauth2Client(client_authn_method=CLIENT_AUTHN_METHOD,
                                            message_factory=CCMessageFactory,
-                                           settings=self._provider_configuration.requests_session)
+                                           settings=self._provider_configuration.client_settings)
 
         provider_metadata = provider_configuration.ensure_provider_metadata(self._client)
         self._client.handle_provider_config(ProviderConfigurationResponse(**provider_metadata.to_dict()),
--- a/src/flask_pyoidc/pyoidc_facade.py
+++ b/src/flask_pyoidc/pyoidc_facade.py
@@ -253,12 +253,8 @@ class PyoidcFacade:
             'token_type_hint': 'access_token'
         }
         logger.info('making token introspection request')
-        token_introspection_request = self._client_extension.construct_TokenIntrospectionRequest(
-            request_args=request_args
-        )
         token_introspection_response = self._client_extension.do_token_introspection(
-            request_args=token_introspection_request,
-            endpoint=self._client.introspection_endpoint)
+            request_args=request_args, endpoint=self._client.introspection_endpoint)
 
         return token_introspection_response
 
--- a/src/flask_pyoidc/pyoidc_facade.py
+++ b/src/flask_pyoidc/pyoidc_facade.py
@@ -4,9 +4,9 @@ import logging
 from oic.extension.client import Client as ClientExtension
 from oic.extension.message import TokenIntrospectionResponse
 from oic.oauth2 import Client as Oauth2Client
-from oic.oauth2.grant import Token
 from oic.oauth2.message import AccessTokenResponse
 from oic.oic import Client
+from oic.oic import Token
 from oic.oic.message import AuthorizationResponse, ProviderConfigurationResponse, RegistrationResponse
 from oic.utils.authn.client import CLIENT_AUTHN_METHOD
 
@@ -55,7 +55,9 @@ class PyoidcFacade:
         """
         self._provider_configuration = provider_configuration
         self._client = Client(client_authn_method=CLIENT_AUTHN_METHOD)
-        self._token = None
+        # Token Introspection is implemented under extension sub-package of
+        # the client in pyoidc.
+        self._client_extension = ClientExtension(client_authn_method=CLIENT_AUTHN_METHOD)
 
         provider_metadata = provider_configuration.ensure_provider_metadata()
         self._client.handle_provider_config(ProviderConfigurationResponse(**provider_metadata.to_dict()),
@@ -143,13 +145,31 @@ class PyoidcFacade:
         Union[AccessTokenResponse, TokenErrorResponse, None]
             The parsed token response, or None if no token request was performed.
         """
+        if not self._client.token_endpoint:
+            return None
+
         request_args = {
             'grant_type': 'authorization_code',
             'code': authorization_code,
             'redirect_uri': self._redirect_uri
         }
+        logger.debug('making token request: %s', request_args)
+        client_auth_method = self._client.registration_response.get('token_endpoint_auth_method',
+                                                                    'client_secret_basic')
+        auth_header = _ClientAuthentication(self._client.client_id, self._client.client_secret)(client_auth_method,
+                                                                                                request_args)
+        token_response = self._client.do_access_token_request(state=state,
+                                                              request_args=request_args,
+                                                              authn_method=client_auth_method,
+                                                              headers=auth_header,
+                                                              endpoint=self._client.token_endpoint
+                                                              )
+        logger.debug(f'received token response: {token_response}')
+        # Store token response in token_class instance. This is required when
+        # making refresh access token request.
+        self._client.token_class = Token(resp=token_response)
 
-        return self._token_request(request_args, state)
+        return token_response
 
     def verify_id_token(self, id_token, auth_request):
         """
@@ -190,46 +210,9 @@ class PyoidcFacade:
                                                                     'client_secret_basic')
         return self._client.do_access_token_refresh(request_args=request_args,
                                                     authn_method=client_auth_method,
-                                                    token=self._token,
+                                                    token=self._client.token_class,
                                                     endpoint=self._client.token_endpoint
                                                     )
-
-    def _token_request(self, request_args: dict, state: str):
-        """Makes a token request.  If the 'token_endpoint' is not configured
-        in the provider metadata, no request will be made.
-
-        Parameters
-        ----------
-        request_args: dict
-            Token request parameters.
-        state: str
-            state is used to keep track of responses to outstanding requests.
-
-        Returns
-        -------
-        token_response: Union[AccessTokenResponse, TokenErrorResponse, None]
-            The parsed token response, or None if no token request was performed.
-        """
-        if not self._client.token_endpoint:
-            return None
-
-        logger.debug('making token request: %s', request_args)
-        client_auth_method = self._client.registration_response.get('token_endpoint_auth_method',
-                                                                    'client_secret_basic')
-        auth_header = _ClientAuthentication(self._client.client_id, self._client.client_secret)(client_auth_method,
-                                                                                                request_args)
-        token_response = self._client.do_access_token_request(state=state,
-                                                              request_args=request_args,
-                                                              authn_method=client_auth_method,
-                                                              headers=auth_header,
-                                                              endpoint=self._client.token_endpoint
-                                                              )
-        logger.debug(f'received token response: {token_response}')
-        # Store token response in Token class instance. This is required when
-        # making refresh access token request.
-        self._token = Token(resp=token_response)
-
-        return token_response
 
     def userinfo_request(self, access_token: str):
         """Retrieves ID token.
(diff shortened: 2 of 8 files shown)


src/flask_pyoidc/message_factory.py
```diff
@@ -0,0 +1,6 @@
+from oic.oauth2.message import AccessTokenResponse, CCAccessTokenRequest, MessageTuple, OauthMessageFactory
+
+
+class CCMessageFactory(OauthMessageFactory):
```
zamzterz, line 4: What is this needed for?
I see the usage of it, but don't understand why/how it's needed 🙂

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -52,9 +55,7 @@ def __init__(self, provider_configuration, redirect_uri):
         """
         self._provider_configuration = provider_configuration
         self._client = Client(client_authn_method=CLIENT_AUTHN_METHOD)
-        # Token Introspection is implemented in extension sub-package of the
-        # client in pyoidc
-        self._client_extension = ClientExtension(client_authn_method=CLIENT_AUTHN_METHOD)
+        self._token = None
```
zamzterz, line 58: Is there any way around having to store this here?
`PyoidcFacade` is intended to be stateless, all data to be persisted is handled in layers above this (and finally ends up in Flask session).
This is making me very wary about this changeset - the statefulness of pyoidc is a bit of pretty big downside going this route... 🤔

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -104,42 +105,51 @@ def authentication_request(self, state, nonce, extra_auth_params):
...
+        response_params: Mapping[str, str]
+            authentication response parameters.
+
+        Returns
+        -------
+        Union[AuthorizationResponse, AuthorizationErrorResponse]
+            The parsed authorization response.
         """
-        auth_resp = self._parse_response(response_params, AuthorizationResponse, AuthorizationErrorResponse)
+        auth_resp = self._client.parse_response(AuthorizationResponse, info=response_params, sformat='dict')
```
zamzterz, line 126: This is the part that will not work without https://github.com/OpenIDC/pyoidc/pull/805
(Just adding note to remember. 🙂 )

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -156,79 +166,94 @@ def verify_id_token(self, id_token, auth_request):
...
         }
+        client_auth_method = self._client.registration_response.get('token_endpoint_auth_method',
+                                                                    'client_secret_basic')
+        return self._client.do_access_token_refresh(request_args=request_args,
+                                                    authn_method=client_auth_method,
+                                                    token=self._token,
+                                                    endpoint=self._client.token_endpoint
+                                                    )
+
+    def _token_request(self, request_args: dict, state: str):
```
zamzterz, line 197: This entire function should be inlined in `exchange_authorization_code` since it's no longer called from any other place.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -238,23 +263,29 @@ def _token_introspection_request(self, access_token: str):
...
-        response = self._provider_configuration.requests_session \
-            .post(self._client.introspection_endpoint, data=request, headers=auth_header) \
-            .json()
-
-        return TokenIntrospectionResponse(**response)
-
-    def client_credentials_grant(self, scope: list = None, **kwargs):
+        # Token Introspection is implemented under extension sub-package of
+        # the client in pyoidc.
+        _client_extension = ClientExtension(client_authn_method=CLIENT_AUTHN_METHOD)
```
zamzterz, line 278: Why recreate this object on every invocation of `_token_introspection_request`?
I'd imagine it more suitable as instance variable (instead of local variable).

zamzterz (commented):
Thanks again for a nice contribution! 🙇 

I realise it's still a draft (pending the changes to pyoidc), but I have some concerns in the meanwhile:
1. This does not address the original intention for manually doing the HTTP requests - to allow users to configure and use a production-appropriate HTTP client. That was the main reason for introducing the custom `request.Session` in `ProviderConfiguration` which has now been completely dropped.
   As mentioned [here](https://github.com/zamzterz/Flask-pyoidc/pull/128#discussion_r805290423) I think the way forward is to combine the approaches (by subclassing `oic.Client`).
1. The statefulness of pyoidc is now leaking to this extension with `PyoidcFacade._token`. I'd really like to see if there's a workaround to avoid that happening. (I haven't yet dug into details why it needs the entire token response for the token refresh...)

src/flask_pyoidc/message_factory.py
infohash: Pyoidc message factories handles request and response classes to parse data respectively. But request and response format vary with the Oauth2.0 flow used. In order to use appropriate classes for the type of data that you are about to send and receive, message factories return the classes that will consume the data. E.g. `oic.oauth2.message.OauthMessageFactory` exists by default.

```python
class OauthMessageFactory(MessageFactory):
    """Factory that knows Oauth2.0 message types."""

    authorization_endpoint = MessageTuple(AuthorizationRequest, AuthorizationResponse)
    token_endpoint = MessageTuple(AccessTokenRequest, AccessTokenResponse)
    refresh_endpoint = MessageTuple(RefreshAccessTokenRequest, AccessTokenResponse)
    # Message as a placeholder ...
    resource_endpoint = MessageTuple(ResourceRequest, Message)
    configuration_endpoint = MessageTuple(Message, ASConfigurationResponse)
```

If you are about to make authorization code flow access token request to the `token_endpoint` of the IdP, `OauthMessageFactory.token_endpoint` will return `AccessTokenRequest` which will construct your request and `AccessTokenResponse` which will parse the response.

But for client credentials flow, the request format is different. So I've created `CCMessageFactory` that overrides `token_endpoint` to handle request and response appropriately. This flexibility allows you to implement custom classes for these formats which then can be passed as a message factory to Pyoidc.

I found this solution in one of their test cases:

https://github.com/OpenIDC/pyoidc/blob/0d316fc0baca21b16164ef17fb275979d178f4b6/tests/test_oauth2.py#L568-L586

infohash:
I love your constructive feedbacks. Your attention to details on code quality has made me more proficient in writing better industrial level code. 🙂

To customize http requests, all delegates have `http_args` argument in which we can pass all parameters as a dictionary supported by `requests` library. Additionally, we can also pass `oic.utils.settings.PyoidcSettings` instance as an argument to `oic.oic.Client` like this:

```python
from oic.oic import Client
from oic.utils.settings import PyoidcSettings
...
self._client = Client(
    ...
    # It takes only 3 arguments.
    settings=PyoidcSettings(
        verify_ssl=False,
        client_cert='cert.pem',
        timeout=10
    )
    ...
)
```

How do I fit `request.Session` in `ProviderConfiguration` here without resorting to deprecate it?

src/flask_pyoidc/message_factory.py
zamzterz: Makes sense, thanks for clarifying! 👍 
(I expected `AccessTokenRequest` to be a generic superclass of possible token request, but I see now it is not - it is only for the "Authorization Code Flow".)

zamzterz:
@infohash I think passing arguments around in all calls (via `http_args`) is bound to become too complex, and as you point out, the `settings: PyoidcSettings` is very limited.

So I was thinking something like this using a subclass:
```python
import copy
from oic.utils.sanitize import sanitize

class ClientWithRequestsSession(oic.oauth2.Client):
    def __init__(self, requests_session):
        # TODO should support all the same constructor args as superclass
        super().__init__()
        self._requests_session = requests_session

    def http_request(self, url, method="GET", **kwargs):
        # All HTTP request are made via this method,
        # so this is the chance to control how the request is made.
        # It has the same behavior as super method except
        # using the requests.Session which already handles cookies automatically
        _kwargs = copy.copy(self.request_args)
        if kwargs:
            _kwargs.update(kwargs)

        if self.req_callback is not None:
            _kwargs = self.req_callback(method, url, **_kwargs)

        try:
            # This is the magic line, instead of `requests.request` use the injected requests.Session instead
            r = self._requests_session.request(method, url, **_kwargs)
        except Exception as err:
            logger.error(
                "http_request failed: %s, url: %s, htargs: %s, method: %s"
                % (err, url, sanitize(_kwargs), method)
            )
            raise

        if self.events is not None:
            self.events.store("HTTP response", r, ref=url)

        return r
```

This class can then be used as a replacement in `PyoidcFacade` with:
```python
self._client = ClientWithRequestsSession(provider_configuration.requests_session, ...)
```

The only caveat is that something similar would need to be done for the "extension client" (`oic.extension.client.Client`) used for making the token introspection requests.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -143,13 +145,31 @@ def exchange_authorization_code(self, authorization_code: str, state: str):
...
+        token_response = self._client.do_access_token_request(state=state,
+                                                              request_args=request_args,
+                                                              authn_method=client_auth_method,
+                                                              headers=auth_header,
+                                                              endpoint=self._client.token_endpoint
+                                                              )
+        logger.debug(f'received token response: {token_response}')
+        # Store token response in token_class instance. This is required when
+        # making refresh access token request.
+        self._client.token_class = Token(resp=token_response)
```
zamzterz, line 170: Unfortunately this doesn't help, it only pushes the problem to another place.
The main problem (which I didn't explain clearly in my last comments) is that this will not work at all given an app where multiple users login.
Since this stores the token here after every token request it will always be overwritten with the last logged-in users tokens.
So imagine the following scenario:
1. User A logs into the app, the token response for User A is now stored.
1. User B logs into the app, the previous token response is replaced with token response for User B.
1. User A hits a scenario where their tokens needs refreshing -> the token response for User B will be used.

That is the reason why everything must be stateless - individual user tokens are stored in Flask session, which by definition is per-user.
Local variables like this are not, they are in a sense "global variables" used for all requests/users of the app.

infohash: Fixed. The only reason I've to use Token class instance for refresh token request is because of this line of code in pyoidc which I'm not understanding why it has to be like this:

https://github.com/OpenIDC/pyoidc/blob/0d316fc0baca21b16164ef17fb275979d178f4b6/src/oic/oauth2/__init__.py#L497-L499

Why overwrite `refresh_token` in `request_args` when I'm already passing refresh_token in request_args. It's only because of this I'm initializing Token like this:

```python
self._client.do_access_token_refresh(request_args=request_args,
                                     authn_method=client_auth_method,
                                     token=Token(resp={'refresh_token': refresh_token}),
                                     endpoint=self._client.token_endpoint
                                     )
```

infohash:
It makes more sense to upstream `request.Session`. I'll raise another PR there to support this.

src/flask_pyoidc/pyoidc_facade.py
zamzterz: Thanks - your new changes LGTM! 👍 

Yeah, I saw the reason for why you did it the original way, but I'm very happy you managed to work around it.
I think the intention of pyoidc, and its reason for overwriting the refresh token, is that it wants to manage all the tokens for you, i.e. it doesn't expect any `refresh_token` in the request_args (which in this use case becomes problematic).

src/flask_pyoidc/pyoidc_facade.py
infohash: Resolved. OpenIDC/pyoidc#805 has been merged.

infohash (commented):
@zamzterz In the latest commit, I have delegated provider discovery response and added support for `requests.Session` (depends on OpenIDC/pyoidc#806). Leaving a note here for you to review.

infohash:
@zamzterz PR is ready to be merged. 🙂

zamzterz:
@infohash Sorry about the delay on this one - I've been really swamped with other work. But I'll review it today! 🎉

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -238,23 +245,22 @@ def _token_introspection_request(self, access_token: str):
...
         Returns
         -------
-        oic.extension.message.TokenIntrospectionResponse
+        TokenIntrospectionResponse
             Response object contains result of the token introspection.
         """
-        request = {
+        request_args = {
             'token': access_token,
+            'client_id': self._client.client_id,
```
zamzterz, line 253: Same here - client credentials shouldn't be sent in the request body.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -294,31 +304,32 @@ def client_credentials_grant(self, scope: list = None, **kwargs):
             auth.clients['default'].client_credentials_grant(
                 scope=['read', 'write'], audience=['client_id1', 'client_id2'])
         """
-        request = {
+        request_args = {
             'grant_type': 'client_credentials',
+            'client_id': self._client.client_id,
```
zamzterz, line 309: Same here - client credentials shouldn't be sent in the request body.

zamzterz (changes requested):
This looks awesome, thanks! 🙇‍♂️ 

I'm wondering about the addition of client credentials in all request bodies (marked in review comments) - as far as I know, pyoidc has the capability to automatically add the client credentials as Basic auth and that should be used instead of the manual way.

One additional thing: `setup.py` should be updated to require `oic>=1.4.0` (if I'm reading correctly that's the latest version containing all your fixes to be compatible with this one).

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -156,79 +185,57 @@ def verify_id_token(self, id_token, auth_request):
...
+        Returns
+        -------
+        Union[AccessTokenResponse, TokenErrorResponse, None]
+            The parsed token response, or None if no token request was performed.
         """
-        request = {
+        request_args = {
             'grant_type': 'refresh_token',
             'refresh_token': refresh_token,
+            'client_id': self._client.client_id,
```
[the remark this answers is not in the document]
infohash: Yep, fixed and slightly redesigned the flow. Same for the others.

zamzterz (approved):
Looking great, thanks once again! 🥇

infohash:
@zamzterz When will be the new release?

zamzterz:
@infohash `v3.11.0` has just been released 🎉

PR #134, opened 2022-04-10
Post authentication function

infohash:
In some environment, it's required to log user sign in and sign out activity. While sign out activity can be logged using logout view function followed by some post logout processing, there's no way to do such post auth processing just after sign in.

The use cases of post auth processing are:

1. To send sign in alerts to the user.
2. To limit user session by one active device only.
3. To send sign in approval request to the user.
4. 2FA.
5. To record location.
6. Custom business operations that involve interacting with the database or with the 3rd party services.

Some of these points can also be covered by IdP itself like 2FA and email notifications but if the user wants to execute its business logic post authentication, this feature provides that functionality.

## Why we should not provide this functionality as part of custom callback function?

Sometimes users accidentally introduce vulnerable code in their business logic. Giving them freedom to override callback function may lead to exposure of secrets not intended by the user. So custom callback function is not an option but to have the ability of post auth execution, this feature is equivalent of having custom callback function.

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -17,7 +17,7 @@ import functools
 import json
 import logging
 import time
-from typing import Optional
+from typing import Any, Callable, Optional
 from urllib.parse import parse_qsl
 
 import flask
@@ -68,6 +68,7 @@ class OIDCAuthentication:
         self.current_token_identity = LocalProxy(lambda: getattr(
             _app_ctx_stack.top, 'current_token_identity', None))
         self._redirect_uri_config = redirect_uri_config
+        self.post_auth_func = None
 
         if app:
             self.init_app(app)
@@ -181,6 +182,8 @@ class OIDCAuthentication:
                                           refresh_token=result.refresh_token)
 
         destination = flask.session.pop('destination')
+        if self.post_auth_func:
+            self.post_auth_func()
         if is_processing_fragment_encoded_response:
             # if the POST request was from the JS page handling fragment encoded responses we need to return
             # the destination URL as the response body
@@ -529,3 +532,22 @@ class OIDCAuthentication:
             return wrapper
 
         return hybrid_decorator
+
+    def post_auth(self, post_auth_func: Callable[[], Any]):
+        """Registers execution of decorated function post authentication.
+
+        Parameters
+        ----------
+        post_auth_func: Callable[[], Any]
+            User defined function that is executed post authentication.
+
+        Examples
+        --------
+        ::
+
+            auth = OIDCAuthentication({'default': provider_config})
+            @auth.post_auth
+            def _post_auth_func():
+                ...
+        """
+        self.post_auth_func = post_auth_func
--- a/tests/test_flask_pyoidc.py
+++ b/tests/test_flask_pyoidc.py
@@ -1038,3 +1038,55 @@ class TestOIDCAuthentication:
         with self.app.test_request_context('/'):
             with pytest.raises(BuildError):
                 authn._get_urls_for_logout_views()
+
+    @patch('time.time')
+    @patch('oic.utils.time_util.utc_time_sans_frac')
+    @responses.activate
+    def test_post_auth(self, time_mock, utc_time_sans_frac_mock):
+        timestamp = time.mktime(datetime(2017, 1, 1).timetuple())
+        time_mock.return_value = timestamp
+        utc_time_sans_frac_mock.return_value = int(timestamp)
+
+        user_id = 'user1'
+        exp_time = 10
+        nonce = 'test_nonce'
+        id_token_claims = {
+            'iss': self.PROVIDER_BASEURL,
+            'aud': [self.CLIENT_ID],
+            'sub': user_id,
+            'exp': int(timestamp) + exp_time,
+            'iat': int(timestamp),
+            'nonce': nonce
+        }
+        id_token_jwt, id_token_signing_key = signed_id_token(id_token_claims)
+        access_token = 'test_access_token'
+        expires_in = 3600
+        token_response = {
+            'access_token': access_token,
+            'expires_in': expires_in,
+            'token_type': 'Bearer',
+            'id_token': id_token_jwt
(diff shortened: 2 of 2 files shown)


zamzterz:
Thanks for a great suggestion! 👍 
Unfortunately, I think this is out-of-scope for this extension - all the use cases listed except 6. (and maybe 2. in some specific edge cases) is something I'd expect the IdP to handle if necessary.

And I suspect there might be some workarounds in the client app, for example using "before request hooks", that would make it possible to achieve without increasing the complexity of this extension.

infohash:
Then how about a decorator for callback function? I disagreed with it initially but flask-oidc also has this.

zamzterz:
I took a quick look at flask-oidc and the [reason stated in the docs](https://flask-oidc.readthedocs.io/en/latest/#custom-callback) mentions using it for keeping the app stateless.
This extension already puts stuff in the session, so statelessness is not achievable by this means.

As for the need for custom callback logic, that can be handled directly in the app code like the following:
```python
def my_auth(f):
    @functools.wraps(f)
    def wrapper(*args, **kwargs):
        session = UserSession(flask.session, PROVIDER_NAME1)
        if session.is_authenticated():
            # we already have a logged-in user

            if flask.session['login_in_progress']:
                del flask.session['login_in_progress']
                print('AFTER auth')
                # TODO user just logged in - do any necessary work
        else:
            # no logged-in user
            flask.session['login_in_progress'] = True

        return auth.oidc_auth(PROVIDER_NAME1)(f)(*args, **kwargs)

    return wrapper

@app.route('/')
@my_auth
def login():
    user_session = UserSession(flask.session)
    return jsonify(access_token=user_session.access_token,
                   id_token=user_session.id_token,
                   userinfo=user_session.userinfo)
```

infohash:
This won't work if I am visiting some other endpoint without session cookie. My approach is to optionally let the user decorate his its function with `@auth.callback` in place of `OIDC_REDIRECT_URI` so that the user can execute its business logic (6th point) every time the callback URL is called for authentication. Why this is out of scope? There should be some place for customization so that the user can hook its own code.

zamzterz:
What do you mean with "visiting some other endpoint without session cookie"? 🙂 

I think it is out of scope since my aim for flask-pyoidc is to be OIDC-compliant. This might mean not being the most flexible/customisable option.
And I agree with your initial reasoning against allowing any custom function as the redirect URI handler - it's seems to be an unnecessary security risk.

infohash:
I have figured out how to achieve custom callback and post authentication without making any changes to flask-pyoidc and it is so easy.

```python
app = Flask(__name__)


def custom_callback():

    print('custom callback execution')
    # TODO - Write your business logic here.
    resp = auth._handle_authentication_response()
    if getattr(resp, 'status_code', None) == 302:
        print('post authentication execution')
        # TODO - Write your post authentication business logic here before
        #  redirecting.
        return resp

    return resp


auth.init_app(app)
# Override the callback view function set by flask_pyoidc.OIDCAuthentication.init_app method.
app.view_functions[auth._redirect_uri_config.endpoint] = custom_callback
```

zamzterz:
1. > if I visit some other view function directly instead of login, the custom callback function will not be executed unless I decorate all views with my_auth.

   Yes, the intention from my example was that the `my_auth` decorator would replace all usages of the regular auth decorator. Any view that does not have either decorator would be unaffected which is expected since they then by definition do not require any logged-in user.

1. Nice, that would also work 👍. Only caveat is that it uses the internal API (`OIDCAuthentication. _handle_authentication_response`) which has no guarantees that it won't change.

1. I do not have any more features on the roadmap right now. The main work that is pending is focused on code improvements - mainly the test code which is way too large and complex to maintain easily.

infohash:
Fixed use of internal API.

```python
app = Flask(__name__)

# Store OIDCAuthentication callback view function to call it internally inside
# custom_callback. If internal API changes, the view function here
# will point to the right callback view function accordingly.
_internal_callback = app.view_functions[auth._redirect_uri_config.endpoint]

def custom_callback():

    print('custom callback execution')
    # TODO - Write your business logic here.
    resp = _internal_callback()
    if getattr(resp, 'status_code', None) == 302:
        print('post authentication execution')
        # TODO - Write your post authentication business logic here before
        #  redirecting.
        return resp

    return resp


auth.init_app(app)
# Override the callback view function set by flask_pyoidc.OIDCAuthentication.init_app method.
app.view_functions[auth._redirect_uri_config.endpoint] = custom_callback
```

Regarding 3, when I was working on test cases for the first time in flask-pyoidc, I found existing test cases so comprehensive that they helped me out pin pointing the exact issues. I can easily maintain all 130 test cases of flask-pyoidc. The next task that I'll be taking is writing type hints and compliance with Numpy docstring format. It seems #132 will take forever to merge 😥.

zamzterz:
Oh, that's a smart workaround for always getting the correct function! 👏 

Well, I'm super glad you found the tests helpful. My main issue with them is there is a lot of code duplication, making it unnecessarily hard to navigate, so something I'd like to improve on.

Type hints would be awesome!

About docstrings - I'm not a fan of the Numpy docstring format.
I initially attempted to follow the [Google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html), but revisiting it again I'd much rather go with the [default Sphinx format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html) which takes much less vertical space (making it easier to get an overview).
What do you think?

infohash:
From the user perspective, numpy docstring has the nicest UI when IDEs render numpy docstring on mouse hover. Numpy doc also has an extension [numpydoc](https://pypi.org/project/numpydoc/) that converts numpy docstring to readme markdown documentation.  Google docstring is just text, no visual appealing and default Sphinx is ugly. Numpydoc is based on Sphinx and follows markdown convention which is rendered by modern IDEs when displaying dostring.

infohash:
From the user perspective numpydoc is better because of markdown convention. From a developer perspective, Sphinx is a reasonable choice. Do you want to finalize Sphinx?

zamzterz:
Yes! 🙂

PR #136, opened 2022-06-05
[RFC-7009] Token Revocation

infohash:
1. To revoke access token or refresh token in case of compromise.
~2. To logout the user from all or unused devices.~
3. To revoke access token obtained by client credentials flow.

## How to use?

```python
auth = OIDCAuthentication({'default': provider_config},
                                      access_token_required=True)
auth.init_app(app)
auth.clients['default'].revoke_token(token='access_token',
    token_type_hint='access_token')
```

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -213,3 +213,13 @@ auth.clients['default'].client_credentials_grant(
 auth.clients['default'].client_credentials_grant(
     scope=['read', 'write'], audience=['client_id1', 'client_id2'])
 ```
+
+## Token Revocation
+If you would like to disable an access or refresh token, simply send a request to the `/revoke` endpoint.
+
+> Note: Revoking a token that is invalid, expired, or already revoked returns a 200 OK status code to prevent any information leaks.
+
+```python
+auth.clients['default'].revoke_token(token='access_token',
+                                     token_type_hint='access_token')
+```
--- a/src/flask_pyoidc/pyoidc_facade.py
+++ b/src/flask_pyoidc/pyoidc_facade.py
@@ -1,4 +1,4 @@
-import base64
+import http
 import logging
 
 from oic.extension.client import Client as ClientExtension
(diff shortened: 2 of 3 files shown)


infohash:
@zamzterz You can release the new version before or after merging this one.

zamzterz:
Logout from all devices is typically managed by the OP (they provide the UI and do the token revocation on their side).
Also, I would not expect that revoking an access token granted to an app would result in logout from all devices - that could be a potential security hole.

infohash:
Oh wait, so we have client credentails

zamzterz:
I would think the app has a continuous need for its client token(s)? And if it by any chance has been compromised, then it would not be a runtime operation to revoke it (more likely something that is done out-of-band in some way).

infohash:
While I agree that token revocation events are extremely rare, the only example that I have is the same as how you can generate the personal access token from your github developer settings and can revoke it anytime. I assume github's own backend service maybe issuing and revoking token at the user's request. Using flask-pyoidc, while issuing token to the user is possible right now with the help of client credentials flow but currently there's no way to revoke it on user's request.

zamzterz:
There's a difference though - a token obtained via client credentials is issued to the client itself, it is not tied to any user.
A common use case is for API auth between services, so I don't really see any case where a user action should influence a client token to be revoked.

infohash:
Closing until some future prospect is there for this feature.

One can still use this feature like this:

```python
auth.init_app(app)
client_auth_method = client._client.registration_response.get('revocation_endpoint_auth_method',
                                                              'client_secret_basic')
client._client_extension.do_token_revocation(
    request_args={token='access_token', token_type_hint='access_token'}, authn_method=client_auth_method,
    endpoint=client._client.revocation_endpoint
)
```

infohash:
@zamzterz So I have figured out how a web service can issue custom access token that is tied to the user requesting it. Using refresh token, you can obtain multiple access tokens. All of them will be active until they are expired so refreshing the token will not revoke the previously issued token which is present in the cookies.

```python
auth.init_app(app)


# Endpoint for the web app to request custom access tokens for developer use.
@app.get('api/issuer')
def issue_custom_access_token_for_developer_apis():
    refreshed_access_token = client.refresh_token(flask.session['refresh_token'])
    # Display it in the web app
    return refreshed_access_token['access_token']
```

Now the user can also revoke them just like in github settings. The main premise of this feature is to bring github's personal access token like functionality in any flask service.

zamzterz:
True, but those tokens would probably be used for something (otherwise, why is the app fetching them?), so if the user revokes it they would break functionality of the app.

The comparison with Github is a bit misleading in this case I think - Github acts as the IdP which goes in line with what I mentioned above; it's typically the IdP's job to provide possibility to revoke tokens for the user.

infohash:
> but those tokens would probably be used for something (otherwise, why is the app fetching them?)

The use case that I am addressing when a user can request multiple access tokens is to use in 3rd party clients to programmatically call the APIs.

> if the user revokes it they would break functionality of the app.

Only issued access tokens can be revoked by the user's request. The web app should only display list of access tokens that have been issued to the user with the time of issue. Along with that, there should be a revoke button that revokes that token on the user's request. It doesn't affect the token in cookies so no functionality of the app will break.

Currently, issuing access tokens and revoking them is still achievable but the backend engineer has to go to extra mile to revoke the token.

zamzterz:
Maybe we're talking about different things with the term "user"?
I'm referring to an "end-user", the person using the web app.
That user can not directly request tokens in any way.
Only the web app can request access tokens on behalf of the user.

Hence, when an end-user has logged in to an app, using for example Google or Github, the user can then visit Google/Github respectively to revoke tokens issued for the web app.
The web app should not display a list of all tokens it has obtained, that would be an anti-pattern.

infohash:
> Only the web app can request access tokens on behalf of the user.

Because the web app can request access tokens on behalf of the user, there should be a way for the user to request the web app to do revocation on his behalf as well which is indirectly is still possible to do from flask-pyoidc by going an extra mile.

**How the web app can do it?**
User can supply the token to revoke that he copied earlier.

zamzterz:
The user should practically never have access to the access token value - no app should give it out back to the user.
That would be the role of the IdP, and then only for "developer" users.

infohash:
How would you issue tokens to 3rd party developers then? They must copy the value to use it in their apps. The issued token must be tied to their user account otherwise you will lose track on who is who. Issuing static API keys for your web app takes away all the good things about using OIDC token.

Besides, we don't have to come up with the application of the feature to provide a utility as long as it's OIDC compliant.

PR #138, opened 2022-06-16
Added caching for token introspection request

infohash:
It is recommended to use caching for token introspection request as there can be many requests fired by the same user. The cached requests must be short lived and time aware. The default size that I have set to expire cache is 5 minutes (300 seconds) and the max cache size is 1024. This can be changed by the user from `ProviderConfiguration`.

The trade off with this is even if the token is revoked for some user, the user will still be able to access the resource for another 5 minutes. It is possible to turn off the caching entirely by setting time to live to zero.

### DIFF
--- a/src/flask_pyoidc/factory/cache_factory.py
+++ b/src/flask_pyoidc/factory/cache_factory.py
@@ -6,7 +6,8 @@ from oic.extension.message import TokenIntrospectionResponse
 
 class TokenIntrospectionCacheFactory(cachetools.TTLCache):
     """Time aware caching factory for token introspection parameters and response."""
-    def __setitem__(self, key, token_introspection_response: TokenIntrospectionResponse) -> None:
+    def __setitem__(self, key, token_introspection_response: TokenIntrospectionResponse,
+                    cache_setitem=cachetools.Cache.__setitem__) -> None:
         """Stores cache of token introspection.
 
         Parameters
@@ -28,6 +29,6 @@ class TokenIntrospectionCacheFactory(cachetools.TTLCache):
             if expires_in_seconds < time_to_live:
                 # Set the attribute to the expiry time of access token.
                 self._TTLCache__ttl = expires_in_seconds
-        super().__setitem__(key, token_introspection_response)
+        super().__setitem__(key, token_introspection_response, cache_setitem)
         # Revert the value of the attribute to the one provided by the user.
         self._TTLCache__ttl = time_to_live
--- a/src/flask_pyoidc/factory/cache_factory.py
+++ b/src/flask_pyoidc/factory/cache_factory.py
@@ -1,0 +1,33 @@
+import time
+
+import cachetools
+from oic.extension.message import TokenIntrospectionResponse
+
+
+class TokenIntrospectionCacheFactory(cachetools.TTLCache):
+    """Time aware caching factory for token introspection parameters and response."""
+    def __setitem__(self, key, token_introspection_response: TokenIntrospectionResponse) -> None:
+        """Stores cache of token introspection.
+
+        Parameters
+        ----------
+        key: cachetools.keys._HashedTuple
+            Token introspection parameters.
+        token_introspection_response: TokenIntrospectionResponse
+
+        Notes
+        -----
+        To maintain user security, the cached access token must be purged at
+        the expiry time of access token or at the expiry time of cache itself
+        or whatever happens first. This method sets custom ttl for the cache,
+        based on what happens first.
+        """
+        time_to_live = self.ttl
+        if token_introspection_response.get('active', False):
+            expires_in_seconds = token_introspection_response['exp'] - time.time()
+            if expires_in_seconds < time_to_live:
+                # Set the attribute to the expiry time of access token.
+                self._TTLCache__ttl = expires_in_seconds
+        super().__setitem__(key, token_introspection_response)
+        # Revert the value of the attribute to the one provided by the user.
+        self._TTLCache__ttl = time_to_live
--- a/src/flask_pyoidc/provider_configuration.py
+++ b/src/flask_pyoidc/provider_configuration.py
@@ -1,10 +1,11 @@
 import collections.abc
 import logging
 
-import cachetools
 import requests
 from oic.oic import Client
 from oic.utils.settings import ClientSettings
+
+from .factory import TokenIntrospectionCacheFactory
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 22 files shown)


infohash:
@zamzterz What do you think? This idea is taken from google documentation on verifying the token. They recommend short lived caching of introspection response.

zamzterz:
Thanks for the PR! 🙇 

I think it falls slightly outside the scope of flask-pyoidc - as you say it impacts the authorisation (the user may have access even after their access token has expired), which I think is best left to the users of this SDK.
It should be easy to the apply caching decorator if needed by the users of this SDK.
But a note in the documentation could possibly added, recommending to do this.

infohash:
Caching can be turned off entirely by setting `ProviderConfiguration.DEFAULT_CACHE_TTL` to 0. By default, we can set it to 0 and let the user to decide if he wants to enable it.

> It should be easy to apply the caching decorator if needed by the users

Like how?

src/flask_pyoidc/provider_configuration.py
```diff
@@ -128,6 +129,8 @@ class ProviderConfiguration:
     """
 
     DEFAULT_REQUEST_TIMEOUT = 5
+    DEFAULT_CACHE_MAXSIZE = 1024
```
zamzterz, line 132: To make it easily configurable these should be added as parameters to the constructor, something like (pseudo-code, needs to fix actual types):
```python
ProviderConfiguration(
  ...,
  introspection_cache_configuration: {max_size: int, max_ttl: int}
```

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -216,6 +216,7 @@ def userinfo_request(self, access_token: str):
 
         return userinfo_response
 
+    @cachetools.cachedmethod(cache=lambda self: self._provider_configuration._cache)
```
zamzterz, line 219: To improve security the caching needs to be done manually: the access token lifetime is returned in response as `exp` so the response should only be cached for `min(self.introspection_cache_configuration['max_ttl'], introspection_response['exp'])`, can this be done with `cachetools.TTLCache`?

src/flask_pyoidc/provider_configuration.py
```diff
@@ -128,6 +129,8 @@ class ProviderConfiguration:
...
+    DEFAULT_CACHE_TTL = 300  # in seconds
```
zamzterz, line 133: Per discussion, default should be `0`.

zamzterz (changes requested):
Thank you! 👏 

The caching needs to be easier to configure and even more restrictive with the TTL for the cache entry to optimise for security.

zamzterz:
> Like how?

I misread the API, it's not easily doable right now, so let's go with your implementation instead.
I've added some review comments to address security concerns.

src/flask_pyoidc/pyoidc_facade.py
infohash: It caches the complete return output of the method. So the full token introspection response is cached otherwise the cached return output will break the caller method if I modify it.

Here's how the cache is being stored:

```python
TTLCache(
    {(<class 'cachetools.keys._HashedTuple'>, 'access_token', 'eyJhbG...'): <oic.extension.message.TokenIntrospectionResponse object at 0x000001ABE53603A0>}, maxsize=128, currsize=1)
```

zamzterz: Yep, good point - that behaviour needs to stay the same.
Still the caching can't rely on decorator, since it only has one globally applied TTL - the TTL needs to be set per response (to respect the `exp`).
I haven't dug into the specifics of `cachetools`, so not sure if it's possible but I think that's a requirement for this change to be introduced.
(Otherwise I'd rather err on the side of caution and simply not cache the response.)

infohash: There are 2 ways of achieving it:

1. In `OIDCAuthentication.introspect_token`, I can write a condition that checks if the `exp` of access token is less than the `ttl` of cache. If it is, then remove the cached token from the cache like this:

https://github.com/zamzterz/Flask-pyoidc/blob/50a67f31944868fbdf90dd2096fc3f82604cde14/src/flask_pyoidc/flask_pyoidc.py#L386-L393

```python
        if result.get('exp', 0) < time.time() + client._provider_configuration._cache.ttl:
            # Pop the cached token if it's about to expire.
            client._token_introspection_request.cache(client).pop(
                client.__token_introspection_request.cache_key(client, access_token=received_access_token))
```

The popping mechanism looks slightly complicated but it works. I'm already using it in `revoke_token` method.

2. Another way is to prevent it from caching in the first place by subclassing the `cachetools.TTLCache` and override the `__setitem__` method to include our condition.

Which one do you think will be more optimal?

infohash: @zamzterz Oh I get it. You want short-lived access tokens to be cached but you want them to be removed from cache if they expire before the cache is expired to prevent unauthorized access. Nice design. I am working on point 2 to see if it can be achieved.

infohash: @zamzterz Done. I hope you like it. 🙂

Now, the access token which is about to be cached will have its own TTL based on `exp`. If the access token will expire before the TTL set by the user, its cache will also be removed. See the new commit.

src/flask_pyoidc/factory/cache_factory.py
```diff
@@ -0,0 +1,34 @@
...
+        the expiry time of access token or at the expiry time of cache itself
+        or whatever happens first. This method sets custom ttl for the cache,
+        based on what happens first.
+        """
+        time_to_live = self.ttl
+        if token_introspection_response.get('active', False):
+            expires_in_seconds = token_introspection_response['exp'] - time.time()
+            if expires_in_seconds < time_to_live:
+                # Set the attribute to the expiry time of access token.
+                self._TTLCache__ttl = expires_in_seconds
```
zamzterz, line 31: This does not look to be safe - it modifies common state of the cache per request and it does it using a private interface.
Unfortunately I don't think this is a suitable solution to this.

zamzterz (commented):
The current implementation of this does not look safe to me (see specific review comment),
and given that it introduces quite a big new functionality for something that can be handled by users of this extension I don't think it makes sense to add this directly in the extension.

infohash:
Closing this as caching can be provided by the framework and is out of scope of OIDC. Also, Cachetools only provides in-memory caching which is not suitable for large-scale services.

PR #139, opened 2022-06-24
Revokes Custom Access Tokens Issued To The User

infohash:
It is possible to issue multiple access tokens to the user by using refresh tokens. These custom access tokens can be used by 3rd party developers to programmatically call the issuer's APIs from 3rd party apps. E.g. A web service may have an endpoint that takes user's request to issue access tokens tied to the user.

```python
auth.init_app(app)


# Endpoint for the web app to request custom access tokens for 
# developer use.
@app.route('/api/issuer')
def issue_custom_access_token_for_developer_use():
    refreshed_access_token = client.refresh_token(
        flask.session['refresh_token'])
    # Display it in the web app
    return refreshed_access_token['access_token']
```

Using token revocation, the user will also be able to revoke them.

```python
client.revoke_token(token='access_token',
    token_type_hint='access_token')
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.9.0
+current_version = 3.10.0
 commit = True
 tag = True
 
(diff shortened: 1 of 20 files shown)


zamzterz:
Related to discussion in #136.

Issue #147, opened 2023-01-19
Verify IDToken fails "Issued in the future"


h4ckd0tm3:
`oic.oic.message.IATError: Issued in the future`

It happens because the time of the token is `1674144487` and `_now` during the check in oic:message.py:800 is `1674144483` so it's of by 4. This would be no problem if `_skew` could somehow be configured in the extension and passed on to verify which supports it via `**kwargs`.

Any different approach I could go?

PR #148, opened 2023-01-19
Add possibilityto to pass **kwargs to pyoidc

h4ckd0tm3:
Would fix Issue #147 and open possibilities for more.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -122,6 +122,7 @@ This extension also uses the following configuration parameters:
   session lifetime (see below). If set to `False` the session will be deleted when the user closes the browser.
 * `PERMANENT_SESSION_LIFETIME`: Control how long a user session is valid, see
   [Flask documentation](https://flask.palletsprojects.com/en/2.0.x/config/#PERMANENT_SESSION_LIFETIME) for more information.
+* `OIDC_CLOCK_SKEW`: Number of microseconds of clock skew allowed when checking the “don’t use before” and “don’t use after” values for tokens.
 
 ### Legacy configuration parameters
 The following parameters have been deprecated:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -122,7 +122,7 @@ This extension also uses the following configuration parameters:
   session lifetime (see below). If set to `False` the session will be deleted when the user closes the browser.
 * `PERMANENT_SESSION_LIFETIME`: Control how long a user session is valid, see
   [Flask documentation](https://flask.palletsprojects.com/en/2.0.x/config/#PERMANENT_SESSION_LIFETIME) for more information.
-* `OIDC_CLOCK_SKEW`: Number of microseconds of clock skew allowed when checking the “don’t use before” and “don’t use after” values for tokens.
+* `OIDC_CLOCK_SKEW`: Number of seconds of clock skew allowed when checking the “don’t use before” and “don’t use after” values for tokens.
 
 ### Legacy configuration parameters
 The following parameters have been deprecated:
--- a/src/flask_pyoidc/auth_response_handler.py
+++ b/src/flask_pyoidc/auth_response_handler.py
@@ -49,11 +49,12 @@ class AuthResponseHandler:
         """
         self._client = client
 
-    def process_auth_response(self, auth_response, auth_request, **kwargs):
+    def process_auth_response(self, auth_response, auth_request, extra_token_args):
         """
         Args:
             auth_response (Union[AuthorizationResponse, AuthorizationErrorResponse]): parsed OIDC auth response
             auth_request (Mapping[str, str]): original OIDC auth request
+            extra_token_args  (Mapping[str, Any]): extra arguments to pass to pyoidc
         Returns:
             AuthenticationResult: All relevant data associated with the authenticated user
         """
@@ -73,7 +74,7 @@ class AuthResponseHandler:
         if 'code' in auth_response:
             token_resp = self._client.exchange_authorization_code(auth_response['code'],
                                                                   auth_response['state'],
-                                                                  **kwargs)
+                                                                  extra_token_args)
             if token_resp:
                 if 'error' in token_resp:
                     raise AuthResponseErrorResponseError(token_resp.to_dict())
--- a/src/flask_pyoidc/auth_response_handler.py
+++ b/src/flask_pyoidc/auth_response_handler.py
@@ -49,7 +49,7 @@ class AuthResponseHandler:
         """
         self._client = client
 
-    def process_auth_response(self, auth_response, auth_request):
+    def process_auth_response(self, auth_response, auth_request, **kwargs):
         """
         Args:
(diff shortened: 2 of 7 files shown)


h4ckd0tm3:
Bump

src/flask_pyoidc/auth_response_handler.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, client):
         """
         self._client = client
 
-    def process_auth_response(self, auth_response, auth_request):
+    def process_auth_response(self, auth_response, auth_request, **kwargs):
```
zamzterz, line 52: Since the `**kwargs` are intended to be forwarded to the token request, I think it makes more sense for this to be a named parameter. Something like `extra_token_params` (with type `Mapping[str, Any]`).
This parameter should also be added to the docstring for this method.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -163,7 +163,10 @@ def _handle_authentication_response(self):
         logger.debug('received authentication response: %s', authn_resp.to_json())
 
         try:
-            result = AuthResponseHandler(client).process_auth_response(authn_resp, auth_request)
+            if "OIDC_CLOCK_SKEW" in current_app.config:
```
zamzterz, line 166: * With above suggestion of `process_auth_response` accepting a dict, this could be simplified:
   ```python
   extra_token_args = {}
   if "OIDC_CLOCK_SKEW" in current_app.config:
       extra_token_args['skew'] = current_app.config['OIDC_CLOCK_SKEW']
   result = AuthResponseHandler(client).process_auth_response(authn_resp, auth_request, extra_token_args)
   ```
* This code base uses single quotes from strings, so it should be `'OIDC_CLOCK_SKEW'`
* This new config option should be documented [here](https://github.com/zamzterz/Flask-pyoidc/blob/main/docs/configuration.md#flask-configuration) and with a new test added for it.

zamzterz (changes requested):
Thank you for the contribution! 💯 

Looks like a solid generalised fix to the #147. 👍 
I've added review comments to make the API of this extension more strict and clear.

h4ckd0tm3:
Will change according to your comments and update the pull request!

h4ckd0tm3:
Implemented the changes + documentation. 
Only thing I'm not really sure about is on where to place the tests.

Should that be in `test_pyoidc_facade.py` or more general in the `test_flask_pyoids.py` OR completely seperated in a new test?

zamzterz:
The new test should go in `test_flask_pyoidc` and assert that the new config value `OIDC_CLOCK_SKEW` is passed on to the token request.

h4ckd0tm3:
Added the test case.

src/flask_pyoidc/flask_pyoidc.py
```diff
@@ -163,7 +163,10 @@ def _handle_authentication_response(self):
         logger.debug('received authentication response: %s', authn_resp.to_json())
 
         try:
-            result = AuthResponseHandler(client).process_auth_response(authn_resp, auth_request)
+            extra_token_args = {}
+            if "OIDC_CLOCK_SKEW" in current_app.config:
```
zamzterz, line 167: Formatting nit: This should also use single quotes `OIDC_CLOCK_SKEW`.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -118,7 +119,7 @@ def parse_authentication_response(self, response_params):
             auth_resp['id_token_jwt'] = response_params['id_token']
         return auth_resp
 
-    def exchange_authorization_code(self, authorization_code: str, state: str):
+    def exchange_authorization_code(self, authorization_code: str, state: str, extra_token_args: [str, Any]):
```
zamzterz, line 122: This should have type `Mapping[str, Any]`.

src/flask_pyoidc/pyoidc_facade.py
```diff
@@ -127,6 +128,8 @@ def exchange_authorization_code(self, authorization_code: str, state: str):
             authorization code issued to client after user authorization
         state: str
             state is used to keep track of responses to outstanding requests.
+        extra_token_args: [str, Any]
```
zamzterz, line 131: Same as above, should be `Mapping[str, Any]`.

tests/test_flask_pyoidc.py
```diff
@@ -11,7 +11,7 @@
 from http.cookies import SimpleCookie
 from jwkest import jws
 from oic.oic import AuthorizationResponse
-from oic.oic.message import IdToken
+from oic.oic.message import IdToken, IATError
```
zamzterz, line 14: Unused import.

docs/configuration.md
```diff
@@ -122,6 +122,7 @@ This extension also uses the following configuration parameters:
   session lifetime (see below). If set to `False` the session will be deleted when the user closes the browser.
 * `PERMANENT_SESSION_LIFETIME`: Control how long a user session is valid, see
   [Flask documentation](https://flask.palletsprojects.com/en/2.0.x/config/#PERMANENT_SESSION_LIFETIME) for more information.
+* `OIDC_CLOCK_SKEW`: Number of microseconds of clock skew allowed when checking the “don’t use before” and “don’t use after” values for tokens.
```
zamzterz, line 125: Is it really "microseconds"? From what I can see it looks like pyoidc accepts skew as number of seconds.

zamzterz (changes requested):
Fantastic!

Some small type issues and a question about the unit of `OIDC_CLOCK_SKEW`.

docs/configuration.md
h4ckd0tm3: Ah yeah, my bad sorry. Got confused digging through the code and finding where it's used. It's indeed seconds!

h4ckd0tm3:
Will push the requested changes tomorrow!

h4ckd0tm3:
Fixed the Issues and Merged upstream. Ready for review.

zamzterz (approved):
Looks great, thanks for finishing this contribution! 👍

h4ckd0tm3:
Will there be a new release with just these changes (and cleanup commits) or will be waited until #151 is closed?

zamzterz:
@h4ckd0tm3 I was holding off for #151, but looks like no new release of pyoidc has landed yet so I've just tagged `v3.13.0` of this extension to include your changes. 🎉

infohash:
@zamzterz Instead of picking up the value of  `skew` from `OIDC_CLOCK_SKEW`, why don't we provide it as an optional argument in `OIDCAuthentication` class? In my view, we should avoid mixing configuration for the extension with user app configuration.

zamzterz:
@infohash Configuration for extensions is commonly specified in the application config: example directly the [Flask docs](https://flask.palletsprojects.com/en/2.3.x/extensions/), so I think it makes sense to keep it as-is.

PR #154, opened 2023-05-02
replace _app_ctx_stack with g for context storage

csfreak:
_app_ctx_stack is deprecated as Flask 2.2.0 and will be removed in Flask 2.4 

see https://github.com/pallets/flask/pull/4682


zamzterz:
Fantastic, looks really good! 👍

PR #159, opened 2023-07-02
Use Front Channel Logout Request For RP-Initiated Logout

infohash:
As we are doing front channel logout to perform RP-Initiated Logout, `pyoidc` has `FrontChannelLogoutRequest` class for it.

### DIFF
--- a/src/flask_pyoidc/flask_pyoidc.py
+++ b/src/flask_pyoidc/flask_pyoidc.py
@@ -12,7 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-
+import http
 import functools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


zamzterz:
Hi @infohash, thanks for looking into this!

The current implementation is following the specification for [RP-initiated logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) at the provider, so the current usage of `EndSessionRequest` (which is tied to the [end_session_endpoint](https://github.com/CZ-NIC/pyoidc/blob/4535d4126d72b40c0518a9856b6e57250012f2bf/src/oic/oic/message.py#L1210)) is correct instead of `FrontChannelLogoutRequest`.

PR #160, opened 2023-07-02
Fix version conflict

infohash:
Pyoidc didn't set version constraint for Pydantic and Pydantic version 2 has been released which has breaking changes that are affecting pyoidc. Until they solve the conflict, add pydantic as a dependency and set its version constraint to less than 2.0

https://github.com/CZ-NIC/pyoidc/issues/861


infohash:
Fixed in upstream https://github.com/CZ-NIC/pyoidc/pull/862

PR #161, opened 2023-07-02
Added JWT Verification In token_auth In decorator

infohash:
Since `token_auth` was introduced, it has been using **[Token Introspection](https://oauth.net/2/token-introspection/)** for JWT verification which is not required. As JWT can be verified in memory, token introspection has now become optional in `token_auth` and `access_control` and is only required for **Opaque tokens**.

`OIDCAuthentication.introspect_token` has been marked for deprecation because its functionality has been merged in `token_auth`.


infohash:
@zamzterz Do you have some time to review?

Issue #164, opened 2023-07-19
Duplicate call to handle_provider_config causing duplicate keys being loaded in keyjar


kerim1:
I encountered a problem where the public keys are loaded twice in the keyjar, causing problems during token validation. I can trace it back to these two calls in the `PyoidcFacade`: https://github.com/zamzterz/Flask-pyoidc/blob/main/src/flask_pyoidc/pyoidc_facade.py#L41-L42
```
provider_metadata = provider_configuration.ensure_provider_metadata(self._client)
self._client.handle_provider_config(ProviderConfigurationResponse(**provider_metadata.to_dict()),
                                    provider_metadata['issuer'])
```
In the first call to `ensure_provider_metadata`, a call to `handle_provider_config` on the client is already made. So why is it necessary to make another call to `handle_provider_config`? The extra call results in keys being added to the keyjar again. My problem seems to be resolved when I remove the second call to `handle_provider_config` without causing any other problems as far as I can see now.

kerim1:
Thanks for the explanation. Now it makes sense to me why the extra call is made. I can also confirm that the issue is resolved on my side once I tried it with your fix in the linked PR.

infohash:
Thanks for confirming. @zamzterz Can you review?

PR #166, opened 2023-07-24
Fixed handle provider config for dynamic discovery

infohash:
See #164


zamzterz:
Looks good, thanks @infohash! 🙇

PR #169, opened 2023-10-11
Fixed bug when server is behind a proxy

GramThanos:
Avoid using the domain and protocol from the request URL, instead use the path and the URL parameters.


infohash:
Nice. This will retain the data in the query params so the user won't have to repeat the request after being redirected by the callback view function.

zamzterz:
@GramThanos Thanks for the great contribution! 👏

Issue #171, opened 2024-02-21
Pass the configuration options in `init_app`


azmeuk:
I encounter issues when initializing flask-pyoidc and using `url_for` to define callbacks URLs.

```python
bp = Blueprint("endpoints", __name__)

user_provider_configuration = ProviderConfiguration(
    ...
    post_logout_redirect_uris=url_for("endpoints.logout", _external=True),
)

auth = OIDCAuthentication({"default": user_provider_configuration}, current_app)

@bp.route("/logout")
@auth.oidc_logout
def logout():
    ...
```

There is a cycling dependency here, the `logout` endpoint needs a `OIDCAuthentication` object, that needs a `ProviderConfiguration`, that calls `url_for`, that needs the endpoint to be initialized.

I think this can be mitigated by delaying the `ProviderConfiguration` part, for instance with the help of `init_app`. I suggest to pass the configuration options to `init_app` so this would be valid:

```python
bp = Blueprint("endpoints", __name__)

auth = OIDCAuthentication()

@bp.route("/logout")
@auth.oidc_logout
def logout():
    ...

user_provider_configuration = ProviderConfiguration(
    ...
    post_logout_redirect_uris=url_for("endpoints.logout", _external=True),
)

auth.init_app(app, {"default": user_provider_configuration})
```

What do you think?

infohash:
`flask.current_app` & `flask.url_for` cannot work outside of [application context](https://flask.palletsprojects.com/en/latest/appcontext/#the-application-context). They can only be called either inside a view function or if the application context is manually pushed otherwise you will get this error:

```shell
RuntimeError: Working outside of application context.

This typically means that you attempted to use functionality that needed
the current application. To solve this, set up an application context
with app.app_context(). See the documentation for more information.
```

So, this line 👇 won't work:
```python
>>> current_app.init_app({"default": user_provider_configuration})
```

You need access to the `app` object here. You can either import `app` object in your module or import your module to the module where your  `app` object lives and register the blueprint there.

```python
...
post_logout_redirect_uris=url_for("endpoints.logout", _external=True),
...
```

This 👆  won't work and not even this 👇:

```python
with app.app_context():
    ...
    post_logout_redirect_uris=url_for("endpoints.logout", _external=True)
    ...
```

This is because in order for the `app` to resolve URL of the view function, the view function must be registered on the app but it hasn't happened yet as the view function has not been initialized yet.

You need to restructure your code:

## `auth.py`
[...]

PR #173, opened 2024-02-29
Lazy Load Provider Config

infohash:
Adds lazy loading of provider config to allow blueprint views to be decorated by auth decorators without initializing auth first. This will provide flexibility to import `auth` instance in blueprint modules before the Flask `app` instance becomes ready.

Fixes:

* Issue: gh-171

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -26,7 +26,14 @@ client_metadata = ClientMetadata(
 provider_config = ProviderConfiguration(issuer='<issuer URL of provider>',
                                         client_metadata=client_metadata)
 
-auth = OIDCAuthentication({'default': provider_config}, app)
+auth = OIDCAuthentication(provider_configurations={'default': provider_config}, app=app)
+```
+
+Flask `app` and provider configurations can be also passed to `init_app`
+to initialize `auth` later.
+```python
+auth = OIDCAuthentication()
+auth.init_app(app=app, provider_configurations={'default': provider_config})
 ```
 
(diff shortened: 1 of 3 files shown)


azmeuk:
I just tested. If endpoints are decorated before the call to `init_app` then a `TypeError` is raised:

```python
endpoints.py:108: in <module>
    @auth.oidc_auth("default")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <flask_pyoidc.flask_pyoidc.OIDCAuthentication object at 0x7a2fdbe79ad0>, provider_name = 'default'

    def oidc_auth(self, provider_name: str):

>       if provider_name not in self._provider_configurations:
E       TypeError: argument of type 'NoneType' is not iterable
```

infohash:
Thanks for noticing. I'll explore a better approach.

infohash:
@azmeuk All it needed was a check to see if  `_provider_configurations` exists or not. Thanks for reporting

azmeuk:
:ok_hand: This works great!

warrenhead:
@zamzterz Any chance of a new release with this PR, to pypi?

